PR #25, opened 2015-01-16
Inheritance: cosmetic improvements and remaining shortcomings

azabluda:
Hello Tony and happy new year!

I finally found some time to look at the shortcomings we have with inherited entities. Problems mentioned in http://trackable.codeplex.com/discussions/552553#PostContent_1278793 and http://trackable.codeplex.com/discussions/552553#PostDetailsCell_1278860 still seem to be valid.

I'm currently working on a more radical solution, but for now would like you to have a look at the following PR which fixes a few obvious glitches in
* TrackableExtensions.CloneObject
* TrackableExtensions.GetChangeTrackingProperty
* TrackableExtensions.GetEquatableMethod
* TrackableExtensions.GetEntityIdentifierProperty

It also reveals the shortcoming in the ModelBase class, which doesn't allow us to call its NotifyPropertyChanged method from the inherited entity (commit ac8f8642cd98c5701ded8f77b6710c01bdb61989).

As I said I'm cooking and shooting out a new concept, which will eventually address not only the ModelBase.NotifyPropertyChanged problem, but generally improve the design. Though for now I suggest that we make just a little step (which this PR actually is) and better understand where TE still falls short.

I advise you to read the commit messages top down to follow the story vs. studying just the final diff :)

Best regards,
AZ

### DIFF
--- a/Source/Tests/TrackableEntities.Client.Tests.Entities/Mocks/MockNorthwind.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests.Entities/Mocks/MockNorthwind.cs
@@ -29,7 +29,7 @@ namespace TrackableEntities.Client.Tests.Entities.Mocks
             Products = new ChangeTrackingCollection<Product>
                 {
                     new Product { ProductId = 1, ProductName = "Chai", UnitPrice = 21.0000M, Discontinued = false, CategoryId = 1, Category = Categories[0] },
-                    new Product { ProductId = 2, ProductName = "Chang", UnitPrice = 20.0000M, Discontinued = false, CategoryId = 1, Category = Categories[0] },
+                    new PromotionalProduct { ProductId = 2, ProductName = "Chang", UnitPrice = 20.0000M, Discontinued = false, CategoryId = 1, Category = Categories[0], PromoCode = "THAI-EXPO" },
                     new Product { ProductId = 3, ProductName = "Aniseed Syrup", UnitPrice = 10.0000M, Discontinued = false, CategoryId = 2, Category = Categories[1] },
                     new Product { ProductId = 4, ProductName = "Chef Anton's Cajun Seasoning", UnitPrice = 22.0000M, Discontinued = false, CategoryId = 2, Category = Categories[1] },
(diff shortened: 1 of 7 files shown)


tonysneed:
Hi @azabluda, Great work so far. Thanks for taking a stab at the inheritance issue.  I read through the commit messages and it looks like a number of things were discovered and resolved.  The major one I noticed back when I looked at this is that the change tracker doesn't play nice with inheritance.  But I think inheritance is an important thing to support - which is sort of obvious. ;-)

I just finished some work and should have a chance next week to work on the the issues I entered into [CodePlex] (https://trackable.codeplex.com/workitem/list/basic) for the 2.5 release.  I'll also transfer those issues over here.

azabluda:
Cool, then it feels like my current work is rather closely related with https://trackable.codeplex.com/workitem/35, https://trackable.codeplex.com/workitem/35 and https://trackable.codeplex.com/workitem/48

Once this PR is merged I will present you my draft.

tonysneed:
Good work.  I took the liberty of added two commits in the pursuit of a possible solution, which is to add an overload of NotifyPropertyChanged to ModelBase which accepts a type argument for the specific type.  However, this is not cast in stone, and we can revert the last commit if your upcoming solution takes us in a different direction.  So just consider it as part of the conversation.

Glad you're working on these issues ... it motivates me to work on the other issues and push toward the 2.5 release milestone.

azabluda:
> overload of NotifyPropertyChanged to ModelBase which accepts a type argument for the specific type

Basically agree. Although one could consider the restriction to TModel type as sort of protection against misuse, it has never been a bullet-proof solution. With the new overload the options to shoot oneself in the foot have just slightly increased.

My new approach (see the upcoming PR) is making the syntax even shorter keeping the security at the same (=low) level
```
NotifyPropertyChanged(() => PromoCode);
```

tonysneed:
Cool, I'll have a look. True what you say about not being bullet proof. I've heard of some possible improvements to the implementation, basically throwing an exception if the property name does not match the property setter from which it is called.=

tonysneed:
But better than exception would be a compiler error or warning. :)

PR #26, opened 2015-01-19
New interface IIdentifiable, new base class EntityBase

azabluda:
This work item addresses the topics raised in http://trackable.codeplex.com/discussions/552553#PostContent_1278793 and http://trackable.codeplex.com/discussions/552553#PostDetailsCell_1278860 as well as in https://trackable.codeplex.com/workitem/35. Has some relationships to #25.

The new interface [IIdentifiable](https://github.com/azabluda/TrackableEntitiesContrib/commit/3cdacbf40de260ed0f97b46a74641d58b64129cd#diff-37b3de12097db9ebc79503f900a9d92bR11) replaces the runtime check https://github.com/azabluda/TrackableEntitiesContrib/commit/3cdacbf40de260ed0f97b46a74641d58b64129cd#diff-fe5e39db73ce5772fb11de64be375c1eL54 with a generic constraint https://github.com/azabluda/TrackableEntitiesContrib/commit/3cdacbf40de260ed0f97b46a74641d58b64129cd#diff-fe5e39db73ce5772fb11de64be375c1eL54 in function MergeChanges.

For backward compatibility I provide a rudimentary implementation of IIdentifiable in ModelBase class (see dfb9ad6). It is neither efficient nor elegant, but essentially just repeats what we always did in the past. I couldn't drop the class entirely or change it any more significantly, because there may still be guys using it in their projects and relying on its specific behavior. Deprecation of ModelBase (should it ever take place) must be more carefully planned and communicated to the users, I believe.

My suggestion for all new TE-based projects is to use the new [EntityBase](https://github.com/azabluda/TrackableEntitiesContrib/blob/e7454fd0fee9876a9e897039c89ab1ad918a4830/Source/TrackableEntities.Client/EntityBase.cs) class which is cleaner, safer and faster due to less private reflection magic. Furthermore EntityBase is non-generic, and thus is less biased towards the root of the entity hierarchy, making the model look somewhat more natural to my taste. Actual entities derived from EntityBase look substantially cleaner: no more unnecessary implementation of IEquatable<T> repeated in every class and subclass, no duplicate implementations of ITrackable, etc. One can just have a look at the commit stats for fb80619f2263d1183c93304da21369c0babfc1a2
>  23 changed files  with 124 additions and *462 deletions*.

Open questions:

1. In fb80619f2263d1183c93304da21369c0babfc1a2 I switched all unit test entities from ModelBase to EntityBase, which effectively reduces the test coverage for ModelBase down to zero. We can of course skip this commit, but then EntityBase remains uncovered. Dilemma... Do you have an idea how to proceed?
2. T4 templates must be updated for cleaner output based on class EntityBase. @tonysneed may I count on your support here?
3. EntityBase overrides Equals/GetHashCode methods, which may cause certain unexpected behavior for those guys who store their entities in HashSets or Dictionaries and one day dare to regenerate classes with the new template. Class [ObjectReferenceEqualityComparer](https://github.com/azabluda/TrackableEntitiesContrib/blob/6cf6ceb37bebbfaecb1e1a6741df468ec262974b/Source/TrackableEntities.Common/ObjectReferenceEqualityComparer.cs) gives a clear answer how to preserve the old behavior, but I guess it must at least be mentioned in the release notes.

Regards,
@azabluda


tonysneed:
Thanks @azabluda for your hard work.  I've been busy with my [own contribution] (https://github.com/aspnet/Mvc/pull/1871) to the asp.net 5 project, but I will definitely process this PR by early next week.  I hope to then work on the tooling part and anything else pending for the v2.5 release.  Cheers!

tonysneed:
I think I'm done with most of my work on the MS contrib - whew! I'll take a look at this next. :smile:

tonysneed:
I looked carefully at your commits and it's a vast improvement.  Thanks! :clap: 

I'll open one or more issues to discuss your questions.

azabluda:
Gladly! :bow:

tonysneed:
@azabluda Is there anything else we need to do to support inheritance? Or does this PR cover it?

azabluda:
Nothing that I know of. Should there arise any further issues, we'd better handle them separately.

Right now I only have a couple of ideas how can the code be slightly beautified, if we completely deprecate ModelBase class. But this isn't urgent and can easily be handled later.

Source/TrackableEntities.Client/ModelBase.cs
```diff
@@ -37,15 +39,128 @@ public abstract class ModelBase<TModel> : INotifyPropertyChanged
...
-        /// <typeparam name="TSpecificModel">Specific entity type</typeparam>
         /// <param name="property">Lambda expression for property</param>
-        protected virtual void NotifyPropertyChanged<TSpecificModel, TResult>
-            (Expression<Func<TSpecificModel, TResult>> property)
+        protected virtual void NotifyPropertyChanged<TResult>
+            (Expression<Func<TModel, TResult>> property)
         {
-            string propertyName = ((MemberExpression)property.Body).Member.Name;
-            if (PropertyChanged != null)
+            NotifyPropertyChanged<TResult>(property);
```
tonysneed: ReSharper complains that line 46 is recursive on all paths. Can we just remove this overload of `NotifyPropertyChanged`?

Source/TrackableEntities.Client/ModelBase.cs
```diff
@@ -37,15 +39,128 @@ public abstract class ModelBase<TModel> : INotifyPropertyChanged
...
+                .SingleOrDefault(m => m.Name == Constants.EquatableMembers.EntityIdentifierProperty);
+            return property;
+        }
+
+        private static class Constants
+        {
+            /// <summary>
+            /// Change-tracking property names.
+            /// </summary>
+            public static class TrackingProperties
```
tonysneed: It looks like `TrackingProperties` isn't used and can be safely removed?

Source/TrackableEntities.Client/ModelBase.cs
tonysneed: Fixed with #37 commit a8bdeb58dd.

PR #42, opened 2015-02-16
[Bug] SetModifiedProperties Should Not Affect Parent or Sibling Entities

tonysneed:
I found and fixed a bug with SetModifiedProperties which has been around a while.  When removing a child entity, ModifiedProperties is cleared from sibling entities because SetModifiedProperties, which is called from RemoveItem in CT collection, sets ModifiedProperties for 1-M and M-M related entities to null.

@azabluda Let me know if you have any thoughts on this.

### DIFF
--- a/Source/Tests/TrackableEntities.Client.Tests.Extensions/TrackableExtensionsTests.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests.Extensions/TrackableExtensionsTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections;
 using System.ComponentModel;
 using System.Linq;
 using System.Collections.Generic;
@@ -7,6 +8,7 @@ using NUnit.Framework;
 using TrackableEntities.Client.Tests.Entities.Mocks;
 using TrackableEntities.Client.Tests.Entities.FamilyModels;
 using TrackableEntities.Client.Tests.Entities.NorthwindModels;
+using TrackableEntities.Common;
 
 namespace TrackableEntities.Client.Tests.Extensions
 {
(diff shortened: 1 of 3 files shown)


azabluda:
Without getting into the details of the problem (neither of the fix)... just a straightforward simplification of the remaining code

```cs
// Prevent endless recursion
ObjectVisitationHelper.EnsureCreated(ref visitationHelper);
if (!visitationHelper.TryVisit(item)) return;

// Iterate entity properties
foreach (var colProp in item.GetNavigationProperties().OfCollectionType<ITrackingCollection>())
{
    // Recursively set modified
    foreach (ITrackable child in colProp.EntityCollection)
    {
        child.SetModifiedProperties(modified, visitationHelper);
        child.ModifiedProperties = modified;
    }
}
```

tonysneed:
Cool thanks.

azabluda:
I will rerun our own tests to see if this change has any negative impact.

tonysneed:
Just checked in the cleaned up code.  Let me know if your tests pass, and I'll merge the PR into dev.

azabluda:
I cherry-picked your 3 commits to our project and all tests are still green :relieved:

azabluda:
Just wondering whether it was intentional or not?

The "failing test" commit belongs to develop, but the "fix" still doesn't.

[image: Inline image 1]

tonysneed:
I thought I pushed the fix.  My local git repo just got corrupted tho ... working on fixing that.

tonysneed:
It looks like the fix got committed to another branch, which I'm about to merge back into dev.

PR #43, opened 2015-02-18
[Bug] DbContext.ApplyChanges Throws Exception when Adding Child M2M Entity

tonysneed:
Added failing test in NorthwindDbContextTests:
Apply_Changes_Should_Mark_Unchanged_Employee_As_Unchanged_And_Added_Territories_Without_Employee_As_Unchanged
- Exception thrown when adding child m2m entity without link back to parent.

Exception takes place when setting relationship state of child entities, because parent entity has not yet been attached to context. This bug takes place in v2.2 but not 2.1.

### DIFF
--- a/Source/Tests/TrackableEntities.EF.5.Tests/NorthwindDbContextTests.cs
+++ b/Source/Tests/TrackableEntities.EF.5.Tests/NorthwindDbContextTests.cs
@@ -1043,6 +1043,51 @@ namespace TrackableEntities.EF5.Tests
 		}
 
         [Test]
+        public void Apply_Changes_Should_Mark_Unchanged_Employee_As_Unchanged_And_Added_Territories_Without_Employee_As_Unchanged()
+        {
+            // Arrange
+            var context = TestsHelper.CreateNorthwindDbContext(CreateNorthwindDbOptions);
+            var nw = new MockNorthwind();
+            var employee = nw.Employees[0];
+            var territory1 = employee.Territories[0];
+            territory1.TrackingState = TrackingState.Added;
+
+            // Causes System.InvalidOperationException:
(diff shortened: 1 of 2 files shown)


tonysneed:
@azabluda If you are able, would you mind please taking a look at this bug? It did not exist in TE v2.1 but appears in v2.2, with refactoring for object graph visitation.  I haven't been able to spot the reason, so I'm hoping you might be able to see it. :smile: I discovered it when running the Relations sample app.  Thanks!

azabluda:
Hi @tonysneed,

It seems to me that the core of the problem has always been there, even prior to PR #7. The old cut off condition for the recursion based on checking the parent type just did not let the algorithm to descend into the affected branch of code. After #7 all branches are visited, which unhides the problem.

Unfortunately I could never fully understand why was 'Added' state processed so differently from the others, even [before](https://github.com/tonysneed/trackable-entities/blob/e6c59e92479ad94fd58efbc63f581c5d0547ef86/Source/TrackableEntities.EF.5/DbContextExtensions.cs#L140) attaching the item itself.

Based on my partial understanding of the processes I may only suggest a very straightforward fix, which just does exactly what the exception message tells us about
```cs
        private static void SetRelationshipState(this DbContext context,
            ITrackable item, ITrackable parent, string propertyName, EntityState entityState)
        {
            // PR #43: Before calling ChangeRelationshipState make sure parent entry is attached
            var parentEntry = context.Entry(parent);
            if (parentEntry.State == EntityState.Detached)
                parentEntry.State = EntityState.Unchanged;

            // Set relationship state for independent association
            var stateManager = ((IObjectContextAdapter)context).ObjectContext.ObjectStateManager;
            stateManager.ChangeRelationshipState(parent, item, propertyName, entityState);
        }
```
It feels "too simple" to my taste, but perhaps it will guide you to a better solution...

Regards,
@azabluda

tonysneed:
I think this should work.  I’ll test and merge.  Thanks!

PR #46, opened 2015-04-10
Unable to place a new order for an existing product

azabluda:
Hi Tony,

Volker has encountered a bug in ApplyChanges yesterday, which I managed to strip down to a relatively simple unit test.

In Northwind terms it is a new Order with one new OrderDetail associated with an existing (unchanged) Product. Exactly how new orders are placed in real life. Strangely ApplyChanges marks Product as Added, which causes EF6 to INSERT it leading to PK index violation.

Could you please look into this problem?

Thanks,
Alexander

P.S: not sure if PR is the right way of submitting problems without any proposed solutions...

### DIFF
--- a/Source/Tests/TrackableEntities.EF.5.Tests/NorthwindDbContextTests.cs
+++ b/Source/Tests/TrackableEntities.EF.5.Tests/NorthwindDbContextTests.cs
@@ -402,6 +402,26 @@ namespace TrackableEntities.EF5.Tests
         }
 
         [Test]
+        public void Apply_Changes_Should_Mark_Unchanged_Product_Of_Added_OrderDetail_Of_Added_Order_As_Unchanged()
+        {
+            // Arrange
+            var context = TestsHelper.CreateNorthwindDbContext(CreateNorthwindDbOptions);
+            var order = new MockNorthwind().Orders[0];
+            var orderDetail = order.OrderDetails[0];
+            var product = orderDetail.Product;
+            order.TrackingState = TrackingState.Added;
+            orderDetail.TrackingState = TrackingState.Added;
+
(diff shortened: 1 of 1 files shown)


tonysneed:
I'll have a look at this and the other PR too.

tonysneed:
I'm able to look at this now -- I had been getting my taxes done and entertaining an out of town guest.  The samples create new orders for existing products, so there might be some other condition.  I'll check out your test.  Thanks!

tonysneed:
It looks like the problem takes place only when the order is a new item.  When the Order is unchanged or modified, then EF does not mark an existing Product entities as Added.  But when the Order is marked as Added, then EF walks down the graph and tries to mark everything as Added, including the Product!  This is actually a well-known idiosyncrasy of EF.

You'll notice that, prior to calling ```ApplyChanges```, the Order property of each OrderDetail is null.  When EntityState is set on an OrderDetail, you'll notice that the Order property is then set.  I'm wondering if we are trying to set the state of Order to Added again, thereby triggering EF to mark downstream entities as Added.  You'll remember we saw the same behavior with PR #43, which dealt with m-m relations.  Would you be able to have a look at the object visitation logic to see if this might be the case?

tonysneed:
I fixed the failing test by updating ApplyChanges to set the correct state for related M-1 entities, correcting the state assignment to Added performed by EF on all downstream entities in an object graph.

azabluda:
Great job! I see Volker's unit test turning green again in his project.

Thank you,
Alexander

tonysneed:
Cool! I’ll push out new NuGet packages - v2.5.1, which include the fixes for #45 and #46.

PR #67, opened 2015-05-14
PortableReflectionHelper

azabluda:
Hi Tony,

Finally got some time for this straightforward code cleanup we discussed a couple of months ago in Skype. I implemented it in two steps for better visibility and traceability (in case one of these commits breaks anything).

Best Regards,
Alex

### DIFF
--- a/Source/Tests/TrackableEntities.Client.Tests.Harness/TrackableEntities.Client.Tests.Harness.csproj
+++ b/Source/Tests/TrackableEntities.Client.Tests.Harness/TrackableEntities.Client.Tests.Harness.csproj
@@ -48,6 +48,9 @@
     <Compile Include="..\..\TrackableEntities.Client\Constants.cs">
       <Link>Constants.cs</Link>
     </Compile>
+    <Compile Include="..\..\TrackableEntities.Client\PortableReflectionHelper.cs">
+      <Link>PortableReflectionHelper.cs</Link>
+    </Compile>
     <Compile Include="..\..\TrackableEntities.Client\TrackableExtensions.cs">
       <Link>TrackableExtensions.cs</Link>
     </Compile>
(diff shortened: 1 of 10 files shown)


tonysneed:
Thanks, I'll have a look at this.  I've been busy releasing v2.5.2 and creating the docs and screencasts.

tonysneed:
Looks good .. I like the idea of an abstract base class so that it's more extensible, allowing for different implementations.  I might have opted for a static class instead, but your approach is more extensible. Although I might have put the conditionals inside of ```PortableReflectionHelper``` to create a different subclass depending on the directives.  But this way is fine too.

azabluda:
Conditional instantiation of an appropriate subclass would be certainly possible, but the problem is that only one subclass out of two is compilable at the time. So we have to decorate the entire Impl with #if..#else..#endif anyway, which makes it sortof redundant to put another #if..#else..#endif in the place of creation.

tonysneed:
I see your point about the number of places where you need conditionals.  I was thinking about the theoretical possibility of possibly having more than one implementation at a time, but that seems like an unlikely possibility.  What about a **static** class then?  Just seems like the singleton might be a little overkill.  Doesn't really matter -- just interested in your thinking. :thought_balloon:

azabluda:
If I understand you correctly, you would have one static class for SILVERLIGHT||NET40 and a similar one for Portable.NET. Both exposing exactly the same static methods with identical signatures, but only one of two being actually compiled.

Foreseeing the trouble that some day one class gets slightly off  its counterpart (and the compiler won't help us identifying it reliably), I intentionally decided to fix the contract very strictly in the abstract class. So now if someone decides to add another method or slightly change an existing signature, he will have to do it first in the abstract class, and the compiler will necessarily warn him if not all Impl classes have been harmonized.

tonysneed:
Ah yes, the versioning angle -- thanks for the exposition. I was actually thinking about a single static class.  But as you point out, the compiler is not able to enforce uniformity between different versions without something like a base class or interface.  Makes sense.

PR #68, opened 2015-05-19
IMergeable interface is not needed?

azabluda:
Hi Tony,

I honestly didn't expect that what we discussed a few weeks ago in Skype would be so easy to implement :blush:

Here is my proposal:

> After some thinking it becomes clear that 'IIdentifiable' contract does not necessarily require
* EntityIdentifier to be a property of an entity. It can be a private field for example.
* EntityIdentifier to be readable and writable by the external code (security issue).
* EntityIdentifier to be of type Guid. Possible alternative in certain N-tier layouts is (interlocked) Integer, or some custom struct with required characteristics (e.g. performance, concurrency, etc.)
* After all, nobody requires this properly to be named 'EntityIdentifier' :)
* One may even think about some non-intrusive ways of cross-tier entity correlation, without any identifier sitting inside the entity.
> 
> EntityBase's implementation of IIdentifiable/ITrackable is still perfectly legit, so let it keep having a public Guid property 'EntityIdentifier' with all necessary serialization attributes. But TE shouldn't force developers to always follow this way. At least there appears to be no strong reason (yet?) to put any constraints on their design decisions regarding the entity correlation techniques.


tonysneed:
Excellent ... I look forward to reviewing this PR. It would be great to have entity correlation to be flexible.

tonysneed:
Yeah, let's get rid of the linkage to IMergeable.  But rather than removing the interface, we should retain it with the ```[Obsolete]``` attribute.  I will then need to remove it from some of the T4 templates and update the samples and VS templates.  So some work will be needed before revving the VSIX.  Let me know if you have any concerns about this.  But I agree that we need to be more flexible here, and that deprecating IMergeable is a good idea.

tonysneed:
I re-added IMergeable for backwards compat, with a note that we'll need to mark it as obsolete after taking it out of T4 templates and VS templates for a new VSIX release.

azabluda:
Ouch, my bad! I always though it should normally be sufficient to just inherit ITrackable (which in turn inherits IMergeable) in the generated entities without explicitly specifying IMergeable in each one. Now I see what I've overlooked https://github.com/tonysneed/trackable-entities/commit/bc1f062186c74e81b4d60ae8b91af7adb19c7e46#diff-5ad62607cb5197877433151f7703c281R46

Yes, fine to mark it as [Obsolete] until the templates/samples are updated. Just to dismiss possible misunderstandings... I would (as always) rely on you maintaining the samples and templates?

tonysneed:
Yes, there is a flavor of entities which implements IMergeable.  I'll be sure to update the templates and samples to remove it.  I don't mind maintaining the samples and templates -- although getting some help would be nice.  The templates and VSIX installers in particular seem to require more time.  I am working on the VS 2015 VSIX installer, but it requires I go back and re-do the project templates to exclude references to packages that are included in the VSIX.  The documentation here is quite sketchy, but something that was strongly suggested for VS2013 is not a requirement for VS2015.

azabluda:
> getting some help would be nice

Anytime! You can always ask my opinion, or simply push your ideas into a Git branch to precess the further discussion.

Issue #69, opened 2015-06-08
Run NUnit in Visual Studio Express


azabluda:
Hi Tony,

Do you have any experience in running TE's NUnit-based tests in Visual Studio Express? The following article worries me a bit http://stackoverflow.com/questions/20907842/how-to-install-nunit-plugin-in-visual-studio-2013-express. Shouldn't we stick to built-in testing framework to encourage the open source community for more active contributions?

Regards,
Alex

tonysneed:
Hi Alex,

I seem to remember you had some issues with NUnit and getting it to work in Visual Studio.  The current "official" testing framework is [xUnit](http://xunit.github.io), which is supported by [Visual Studio](http://xunit.github.io/docs/running-tests-in-vs.html).  It would be great to migrate the tests from NUnit to xUnit.

In terms of Visual Studio editions, be aware that VS Express is now legacy.  Instead developers need to use [Visual Studio Community Edition](https://www.visualstudio.com/en-us/products/visual-studio-community-vs.aspx), which is basically a free version of VS Professional, and it supports running xUnit tests from the built-in Test Explorer. :smile: 

Cheers,
Tony

azabluda:
Thanks for the info! I will try to migrate our tests to xUnit, though I might need your help with SpecFlow tests.

tonysneed:
Terrific, thanks! I'll let you know about the SpecFlow tests.

tonysneed:
Closing this issue as implemented in #70.

PR #70, opened 2015-06-09
Migrate from NUnit to xUnit.net

azabluda:
Hi Tony,

Here comes my proposal for #69. As always I tried to split my work into smaller chunks to make things more comprehensible.

Commits b50c061, 412acd3 and 4684217 are more or less machine generated and don't require any thorough review. So please have a look at the remaining revisions. I find some of them (e.g. 87b2a44 or b8cb3c3) rather peculiar.

Best regards,
Alex


tonysneed:
The migration to xUnit.net looks good.  I see you even updated the SpecFlow tests -- thanks!  It's something I've wanted to have done for a long time, so I'm glad you took it on.  All the tests show up in the R# test runner too. :smile:

Issue #75, opened 2015-07-15
GetChanges can fail to remove restored deletes for items in multiple ChangeTrackingCollections


stevedelong:
In a similar vein to the issue I just submitted about deletion of items that belong to multiple ChangeTrackingCollections, I've found a case where GetChanges can leave restored deletes in collections because of where the visitation is checked.

In TrackableExtensions.RemoveRestoredDeletes, the visitation helper is checked before the item is removed from the collection, which works perfectly fine in single-parent cases, but when an item is part of more than one collection, it will only be removed from the first one visited, at which point all others will trip the visitation check and continue prior to being removed from the collection.

In my testing, moving the item removal from after the navigation property iteration to just above the visitation check takes care of this, but I'm not sure what else that might affect.

I have a test case that illustrates the removal failure, but as I said in my other issue, I'm not sure how best to submit it to you.

tonysneed:
Thanks for submitting this issue. Yes, it would help to have a failing unit test. The best way is to fork the repo, create a local branch, then add a unit test to the appropriate test project. Then commit your changes and publish the branch back to GitHub. If you go to the web site you'll then see a button called Compare and Merge (I think). Clicking that button should give you the option to submit a pull request, from which you can reference this issue. Once you've done it a couple times, submitting pull requests will be second nature.

stevedelong:
I’ve submitted a pull request with test cases for the two issues – hopefully I’ve done it properly in a useful way, this is my first-ever GitHub pull request (and in fact, my first GitHub contribution in any form whatsoever).

I’m out of town until 7/20, but will at least try to keep an eye on the tracker until then.

tonysneed:
I'm also on vacation until July 20th, but I'll see if I can have a look at the PR after the kids are in bed. ;)

Congrats on your first pull request and Github contrib!!

tonysneed:
Resolved by working with each change tracker independently.  Refactored tests to pass in commit 8e66ad282518ea7618148441ce8ee1d4405e38c9 for pull request #92 [Workarounds for Issues 74-75].

Issue #74, opened 2015-07-15
Removing an item with TrackingState.Added from ChangeTrackingCollection possible incorrect behavior


stevedelong:
I've found an issue when deleting a new item added to a ChangeTrackingCollection.  The sequence of events is as follows:

1) CTC has tracking set to true.
2) New item is created, added to CTC, its state gets set to TrackingState.Added as expected.
3) CTC.Remove(newItem) is called.  Item's state is set to Unchanged, and is added to the CTC's _deletedEntities collection.

Ordinarily this isn't a problem, as GetChanges won't care about it because it's Unchanged.  However, if the item belongs to multiple collections, deleting it means removing it from all of them, which changes the steps to be:

1) CTC1 and CTC2 both have tracking set to true.
2) New item is created, added to CTC1, its state gets set to TrackingState.Added as expected.
3) Item is added to CTC2, its state is already TrackingState.Added and is not changed.
4) CTC1.Remove(newItem) is called.  Item's state is set to Unchanged, and is added to CTC1's _deletedEntities collection.
5) CTC2.Remove(newItem) is called.  Item's state is Unchanged, and is now set to Deleted.  Item is added to CTC2's _deletedEntities collection as well.

If I call GetChanges on either CTC1 or CTC2, I get newItem as a Deleted item, which then fails on an attempt to save because it doesn't actually exist.

My thoughts on a fix for this are to modify TrackableExtensions.SetState so that if it's setting state to Deleted it simply leaves an Added item's state alone vs setting it to Unchanged.  ChangeTrackingCollection.RemoveItem even looks for Added items when determining whether an item should be added to the _deletedEntities collection.

I have a test case that illustrates this issue, but am not sure how best to submit it for you to review, so let me know what you'd like for that.

tonysneed:
Resolved by working with each change tracker independently.  Refactored tests to pass in commit 8e66ad282518ea7618148441ce8ee1d4405e38c9 for pull request #92 [Workarounds for Issues 74-75].

PR #76, opened 2015-07-16
Added test cases demonstrating issues #74 and #75

stevedelong:
Added two new test cases (and a trivial multi-parent model) model:

Removed_Added_Items_With_Multiple_Parents_Should_Not_Be_Marked_As_Deleted_Or_Modified
for issue #74

GetChanges_Should_Not_Leave_Restored_Deleted_Items_With_Multiple_Parents_In_Source_Graph
for issue #75

I've not yet added proposed changes, but as a quick summary of my thoughts:

#74: TrackableEntities.cs, SetState.  For setting a state of Deleted for an Added item, don't modify its state so that it can be removed from multiple collections using the same logic.  ChangeTrackingCollection.cs, RemoveItem is expecting a TrackingState.Added to not add to _deletedEntities already anyway.

#75: TrackableEntities.cs, RemoveRestoredDeletes.  Move the "remove item if marked as deleted" block above the "Prevent endless recursion" line, so that the item is always removed from its parent collection regardless of recursion.

### DIFF
--- a/Source/Tests/TrackableEntities.Client.Tests.Entities/MultiParentModels/MultiParentChild.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests.Entities/MultiParentModels/MultiParentChild.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace TrackableEntities.Client.Tests.Entities.MultiParentModels
+{
+	public class MultiParentChild : EntityBase
+	{
+	}
+}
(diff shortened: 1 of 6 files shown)


tonysneed:
@stevedelong,

I am just now circling back to some outstanding issues, so thanks for being extremely patient. :smile: 

These issues pertain to **multiple change trackers** updating state on an entity which is shared among them.  This scenario is bound to result in conflicts, because one change tracker is not aware of state changes brought about by another change tracker.  This is the result of a design decision of using change tracking collections, rather than a single all-knowing change tracker.

The way to get around this limitation with regard to the **children with multiple parents scenario** is work with change trackers *independently*, either by grouping operations or by using more than one change tracker.

See my comments for issues #74 and #75 for examples of how to do this.  I have created pull request #92 [Workarounds for Issues 74-75].

Note: We may change this architecture in a future version of Trackable Entities that is targeted to Entity Framework 7 or later, when we'll likely leverage improved change-tracking facilities of EF7 and use the single change-tracker approach.

Look at commit 8e66ad282518ea7618148441ce8ee1d4405e38c9 to see how I refactored your tests to get them to pass by working with multiple change trackers independently: `root.Parent1Items1` and `root.Parent2Items`.

tonysneed:
Closing the PR and issues.  @stevedelong, please feel free to comment on any of them.  I can re-open if needed.  Cheers.

PR #92, opened 2015-12-09
Workarounds for Issues #74-75

tonysneed:
Updated the following tests to illustrate workarounds with children that have multiple parents:

1) Removed_Added_Items_With_Multiple_Parents_Should_Not_Be_Marked_As_Deleted_Or_Modified
2) GetChanges_Should_Not_Leave_Restored_Deleted_Items_With_Multiple_Parents_In_Source_Graph

These tests now both pass as a result of working with multiple change trackers independently.  This is required in order to avoid conflicts among multiple change trackers operating on the same entities.  The reason for this is that Trackable Entities uses change tracking collections rather than a single all-knowing change tracker.  We may change this architecture in a future version of Trackable Entities that is targeted to Entity Framework 7 or later, when we'll likely leverage improved change-tracking facilities of EF7 and use the single change-tracker approach.

### DIFF
--- a/Source/Tests/TrackableEntities.Client.Tests.Entities/MultiParentModels/MultiParentChild.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests.Entities/MultiParentModels/MultiParentChild.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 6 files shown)


tonysneed:
Closing the PR and issues, as the scenario of entities tracked by more than one change tracker is not supported.  I created this PR to demonstrate workarounds.

Issue #80, opened 2015-09-12
LoadRelatedEntities() - missing table in generated SQL


luboshl:
Hi Tony,

first thank you for your Trackable Entities :)

I have problem with DbContextExtensions.LoadRelatedEntities() method when I use entity which references other entity with base class. (I uploaded full example to https://dl.dropboxusercontent.com/u/91293338/TrackableEntities/Demo.zip)

[Table("LightOnActions")]
public class LightOnAction : ActionBase
{
  public int? Light_Id { get; set; }

  [Required]
  [ForeignKey("Light_Id")]
  public LightDevice Light { get; set; }
}

[Table("LightDevices")]
public class LightDevice : DeviceBase { ... }

[Table("Actions")]
public abstract class ActionBase : EntityBase { ... }

When I use LoadRelatedEntities, I get System.Data.Entity.Core.EntitySqlException, because the generated SQL is "SELECT VALUE x FROM  AS x WHERE x.Id IN {1}" - table is missing.

Example:

using (var context = new DemoContext())
{
  var action = new LightOnAction
  {
    ...
    Light = new LightDevice
    {
      ...
      TrackingState = TrackingState.Added
    },
    TrackingState = TrackingState.Added
  };

  context.ApplyChanges(action);
  context.SaveChanges();
  
  // here I get the Exception
  context.LoadRelatedEntities(action);
  action.AcceptChanges();
}

Is there a way to solve the problem in this case?

tonysneed:
Hello @luboshl,

I'm worry it's taken me to long to respond.  Been crazy busy until now.  I've taken a look at your demo and stepped through the code.  It seems that `LoadRelatedEntities` probably has a problem with FK values which are **nullable**.  I'll see if I can replicate this in a small failing test.

In the meantime, were you able to find a workaround?

Cheers,
Tony

tonysneed:
I'm still trying to reproduce the error.  I have tried nullable foreign keys, and also reference types which inherit from a base class, but so far no errors.  Next, I'll try multi-level reference types, in order to approximate the demo which @luboshl updloaded.

tonysneed:
I was finally able to locate the source of the problem, then I produced a failing test to replicate it, and I fixed the problem with PR #93: Load Related Entities for Derived Reference Types.

luboshl:
Thank you, Tony!

PR #93, opened 2015-12-15
Load Related Entities for Derived Reference Types

tonysneed:
When a type has a reference type which derives from a bass class, then DbContext.LoadRelatedEntities throws an exception because there is a missing table name in the resulting Entity SQL.

I have corrected the problem with this pull request and included a test which verifies the fix.


tonysneed:
This PR fixes issue #80: LoadRelatedEntities() - missing table in generated SQL.

Issue #105, opened 2016-01-16
ITrackableExtended<TProperty> interface


tonysneed:
We should create an `ITrackableExtended<TProperty>` interface in a PCL called TrackableEntities.Common.Extended, which references TrackableEntities.Common.  The interface should inherit from `ITrackable` but add a `TrackgingStateExtended` property of type `<TPropertyType>`.

This provides an extensibility hook for supplying additional state metadata, which can be utilized for extending the `DbContext.ApplyChanges` method in the server library.  An example would be for specifying that entities on one side of a M-M relation are to be created rather than merely added to the relationship.

This issue was first proposed in #102 and discussed in this [comment](https://github.com/TrackableEntities/trackable-entities/issues/102#issuecomment-171965676).

luboshl:
Ok :smile: What about namespace? `TrackableEntities.Common.Extended` or just `TrackableEntities.Extended` as `TrackableEntities` namespace for `ITrackable`?

And what exactly does "PCL" mean? I understand it as VS project, but what is that abbreviation for?

tonysneed:
The namespace should be the same as the project: TrackableEntities.Common.Extended

PCL stands for Portable Class Library.

tonysneed:
Sorry, I think we made an exception to this namespace rule for ITrackable, so we should now use: TrackableEntities.Extended

luboshl:
@tonysneed, this issue can be closed.

PR #108, opened 2016-01-16
Added ITrackableExtended interface.

luboshl:
Created on the basis of **#105: ITrackableExtended&lt;TProperty&gt; interface**


luboshl:
Wow, that is not what I expected :( I created new local branch "extended-states", in its history i can see only one commit: https://github.com/luboshl/trackable-entities/commit/a5553d4216b2b910079340129a0ddd780da6c24c, but here I can see more commits including reverts... :(

I think final changes are what it should be, but I haven't expected other commits. You can reject this PL and I can try to do it better :disappointed:

tonysneed:
I'll have a look at this -- thank you!

tonysneed:
No worries - after I review it, I'll have you do a squash commit, so it will be just one commit in the end.

luboshl:
I've created ne PR #110.

Issue #106, opened 2016-01-16
DbContext.ApplyChanges overload accepting lambda


tonysneed:
We should add an overload to the `DbContext.ApplyChanges` method which accepts a lambda expression, in which the caller could apply logic for setting `EntityState`.  This would allow, for example, the caller to inspect the tracking state or extended tracking state of the entity and then apply some logic for how the `EntityState` should be set.

For example, the caller could determine that, if the [extended tracking state](https://github.com/TrackableEntities/trackable-entities/issues/105) of an entity has been set to `Created` and the entity is being added to an M-M relation, then the `EntityState` should be set to `Added`.  Otherwise, it can be set to `Unchanged`, which will cause it to be added to the M-M relation as an existing entity.

Another possibility would be for the called to check to see if the primary key property of the entity is an identity and the value is zero, and it is in an M-M relation, then the state should be set to `Added` versus `Unchanged`.

This issue was first proposed in #102 and discussed in this [comment](https://github.com/TrackableEntities/trackable-entities/issues/102#issuecomment-170888322).

luboshl:
I started to implement this, but I encountered a problem how to implment that. There is no problem to add overloades for `DbContext.ApplyChanges` with func returning state for given item: `Func<ITrackable, EntityState>`. But what next? If caller implements some logic with some result state, how to use it in the private implementation of `ApplyChanges`? There are so many places where `context.Entry(item).State` is set depending on different conditions...

I created first sample in commit https://github.com/luboshl/trackable-entities/commit/69b586bebe03173bd07ea4fb3dfd7315358504a3

But I think there should also be abillity to set the relation in `context.SetRelationshipState`. The more I think about it the more I mean there must be something more, but I am not sure what...

tonysneed:
I will take a look at this today. Thanks!

tonysneed:
Before I have a look at your sample, let's discuss a general approach, based on your use case of properly applying changes to entities in a M-M relation.  Perhaps we can then write some tests prior to writing an implementation (TDD). :smile:

[...]

PR #111, opened 2016-01-17
Created DbContext.ApplyChanges overloads accepting lambda

luboshl:
Based on  #106

### DIFF
--- a/Source/Tests/TrackableEntities.EF.5.Tests/NorthwindDbContextTests.cs
+++ b/Source/Tests/TrackableEntities.EF.5.Tests/NorthwindDbContextTests.cs
@@ -1925,6 +1925,113 @@ namespace TrackableEntities.EF5.Tests
 			Assert.Null(order.CustomerId);
 		}
 
-		#endregion
-	}
+        #endregion
+
+        #region Apply Changes with 'state selector' callback
+
+        [Fact]
+        public void Apply_Changes_With_State_Selector_Should_Mark_Single_Entity_Added()
+        {
+            // Arrange
+            var context = TestsHelper.CreateNorthwindDbContext(CreateNorthwindDbOptions);
+            var product = new Product();
+            product.TrackingState = TrackingState.Unchanged;
+
+            // Act
+            context.ApplyChanges(product, (e, rs) =>
+            {
+                // no matter what 'e' and 'rs' is set to,
+                // set state to 'Added'
+                return EntityState.Added;
+            });
+
+            // Assert
+            Assert.Equal(EntityState.Added, context.Entry(product).State);
+        }
+
+        [Fact]
+        public void Apply_Changes_With_State_Selector_Should_Mark_Single_Entity_Unchanged()
+        {
+            // Arrange
+            var context = TestsHelper.CreateNorthwindDbContext(CreateNorthwindDbOptions);
+            var product = new Product();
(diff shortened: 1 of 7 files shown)


tonysneed:
Thanks @luboshl for the PR.  I'm having a look at it now.

I can see an issue with the method signature for the new overload of `ApplyChanges`, which explains why you were having to cast `e` as you mentioned in your [comment](https://github.com/TrackableEntities/trackable-entities/issues/106#issuecomment-172392574).

I will start with comments on your first commit: 1fc8527ed69b15ceb5de1899882c8ac820a1dff7. Heading over to do that now.

tonysneed:
By the way, I am located in Europe, which is GMT +1.  So I am now starting my day.  Where are you physically located?

It is also possible for us to chat in real time.  I created a chat room for this repo on [gitter](https://gitter.im).  Let me know if you are able to join.

luboshl:
Yes, exactly. But it's question, if it is better to have two or more collback (each for different type) or to have only one and if I need to use it for different types, then I have to cast it:
```C#
context.ApplyChanges<ITrackable>(order, (e, rt) =>
{
    var orderEntity = e as Order;
    var orderDetailEntity = e as OrderDetail;

    if (orderEntity != null)
        return ...

    if (orderDetailEntity != null)
        return ...
}
```

luboshl:
But you are not able to do this:
```C#
public static void ApplyChanges<TEntity>(this DbContext context, ITrackable item,
    params Func<TEntity, RelationshipType, EntityState?>[] stateSelector)
```
because you have only one `TEntity` (for example `Order`), but you want to have two callback - for `Order` and for `OrderDetail` as input argument in the callback function
In this case you still need to set `TEntity` to `ITrackable` (or to any base class, if exists) an then cast it to concrete type.

tonysneed:
This PR is looking good so far.  We may need to add a couple more tests.

But then we should be able to implement the two new `ApplyChanges` overloads.  We will need to add a `stateSelector` parameter to the private `ApplyChanges` method as an optional parameter defaulting to `null`.  If it is not null, we invoke it, then if it does not return null, we set the entity state to whatever is returned.

luboshl:
I expected we will create more test. Different relationships and combinations of states...

I think we should add `IList<IInterceptor>` collection, one we have created this interface and method `AddInterceptor` :) But it is true, that it is not absolutely necessary. We have only one implementation of `IInterceptor` so we can add its `stateSelector` directly. In the future, if we add some more, only internal implementation (I mean private `ApplyChanges` method) will change. Still if we want to use `stateSelector` directly, we have to put collection of them.

tonysneed:
Yes, we will probably need to refactor the private `ApplyChanges` method when we implement #114.  But for this PR, we can go ahead and just add an optional `stateSelector` parameter to the private `ApplyChanges` method.  Then when we go to implement #114, we can [refactor](https://en.wikipedia.org/wiki/Test-driven_development#5._Refactor_code).  Because of the complexity of change tracking, I've found it beneficial to adopt a TDD mentality. :smile:

luboshl:
I meant we **are** in the other PR (#114) and that this (#111) is deprecated :)

tonysneed:
Ah, I see now that you merged your 105 branch into the 106 branch.  So we can close this PR and proceed with #114.  That will make it easier.

I'm now looking at that PR and I can make commits on your commits there.

PR #113, opened 2016-01-18
Temporary PR to Demonstrate Fluent API interception

tonysneed:
Sample implementation.


tonysneed:
We're done with this now.

PR #114, opened 2016-01-18
Added intenceptors supportting fluent API and concrete StateChangeInterceptor

luboshl:
Based on issue #106


luboshl:
A note to added `ApplyChanges` overloads - they are both `internal`, only to be called from `InterceptorExtensions` in fluent. They are not meant to be called from outside. Or do you want also to add another `ApplyChanges`, that allows a caller to add lambda directly to this method (as we planned in the beginning)? My personal opinion is not to combine fluent and direct call to `ApplyChanges`. What do you mean?

luboshl:
I also made *squash* of commits, so it's much cleaner now. I know that it doesn't have to be the final solution, but I read the process again and wanted to try complete procedure including squashing. It was little complicated because there was a conflict, but I resolved it successfully :)

tonysneed:
I will have a look at this today and respond with comments.  Thanks!

tonysneed:
Alright, I added all my comments in a3c80766bd755d208c41b3ed32ba474bcd41babc.

You did a great job -- thanks!  And I was pleased to see all the xml-doc comments supplied. :smile: 

The main changes are:
- `IInterceptor` has a `GetEntityState` state method, so that `ApplyChanges` has a method to invoke.
- `StateChangeInterceptor` is storing the selector as a `Delegate` and dynamically invoking it.  We'll need to set the selector with a method rather than a ctor because of the `TEntity` type argument.
- We need to move the `ApplyChanges` methods from `InterceptorExtensions` to `DbContextExtensions`, where they belong.
- I also added an additional `ApplyChanges` overload which accepts a `stateSelector`.

Let me know if these changes make sense.  Cheers!

azabluda:
Only little remark from my side ragarding the `AddXxxInterceptor` method.

Term `Add` hints towards mutability and thus one may expect the following code to work:

```cs
var dbCtx = CreateNorthwindDbContext(...);
dbCtx.AddStateChangeInterceptor(someFunc);
dbCtx.ApplyChanges(...); // Why is my interceptor not used?!
```

In this respect I like the `With` prefix much better, because it implies immutability. Unfortunately the current implementation is only half way immutable. The clean solution should allow the following:

```cs
var dbCtx = CreateNorthwindDbContext(...);
var xCtx = dbCtx.WithStateChangeInterceptor(x);
var xyCtx = xCtx.WithStateChangeInterceptor(y);
var xzCtx = xCtx.WithStateChangeInterceptor(z);

// Both of the following lines use all three interceptors 'x', 'y' and 'z'
xyCtx.ApplyChanges(...); // Expected: only 'x' and 'y' interceptors are used
xzCtx.ApplyChanges(...); // Expected: only 'x' and 'z' interceptors are used
```

luboshl:
I thik we missed in one important thing - I meant `IInterceptor` to be realy common interceptor for any action. Now we are implementig one concrete interceptor - `StateChangeInterceptor`. This interceptor explicitly sets the state of the entity no matter what standard process is.

But what I understand from your previous comments, you meant `IInterceptor` to be interface for *any interceptor that changes state of the entity*. It means now we have interceptor `StateChangeInterceptor` which changes entity state based on relationship and the entity itself (input arguments in `Func<>`) and in the future any other interceptor implementing `IInterceptor` will somehow change state of the entity (for example based on state of some process or whatever).

That's wery important to be clear, because in that case it makes sense to have method `GetEntityState` in the interface `IInterceptor`, bot not in the first case. That's why for now I made `IInterceptor` empty without that method.

So what do you think is the perspective of `IInterceptor`? If it is only for changing state, let's then rename it to `IStateInterceptor` and give him method `GetEntityState`. If not, then let's make him empty.

It's true, that absolutely main porpose (maybe the only one) of the `ApplyChanges` method is to set state to a graph of entities. But I can imagine other interceptor. For example some, that indicates whether to go through all descendant of given entity or not. And this interceptor would have for example this method:

```C#
internal bool ShouldApplyChangesOnProperties(ITrackable item)
{
    var entity = item as TEntity;
    return entity != null ? this.predicate(entity) : true;
}
```
and could be used for example as this:
```C#
context
    .AddShouldApplyChangesOnPropertiesInterceptor<Order>(e => e.OrderId == 1)
    .ApplyChanges(order);
```

That's realy only fictional example. But on this example you can see, that each interceptor needs other method. This is how I understand the `IInterceptor` and all the stuff around :smile: 

We have to clarify the core of all this before I continue and apply your comments.

tonysneed:
Yes I agree, we should use the With prefix instead of Add.

tonysneed:
Good suggestion to rename IInterceptor to IStateInterceptor. Let's so that and if possible proceed.

I can respond in more detail a bit later, but the interface provides a contract with ApplyChanges, and the caller can achieve what you are suggestion within this constraint. More detail to follow if clarification is needed ...

tonysneed:
If we were to have an IInterceptor interface that has no methods, how would it be called by ApplyChanges?

I feel that the generic method (I think it's called GetEntityState) supplies the caller with what he needs to decide what state to return.  But let me know if you think more or different parameters are needed. And within the callback, the caller can do anything they want to determine which state to return, for example, some lookup in a cache. Seems fairly flexible to me, given the kind of extensibility we are allowing, which is to allow external code is to determine what state to set the entity to, using a strategy pattern and fluent API.

luboshl:
Yes, and that's the point - do we only allow with intenceptor (any, in the future, not only this current `StateChangeInterceptor`) to change entity state? See my example above (https://github.com/TrackableEntities/trackable-entities/pull/114#issuecomment-172834269) with an interceptor allowing to decide whether to iterate throuch descendant or not. That interceptor **doesn't change state** of any entity. So that interceptor doesn't need method `GetEntityState`. It needs `ShouldApplyChangesOnProperties` (which returns `bool`) from the example above.

luboshl:
Do you understand my example of `ShouldApplyChangesOnPropertiesInterceptor`? If not, I can write it in the other way :smile:

tonysneed:
I have to look at your example mor closely -- I'm out of the house and responding on my phone. :)

luboshl:
Ok, it can wait, of course. I was ill last few days, so I also had more time and could respond faster. Now I am back at work so I it can wait :smile:

luboshl:
@azabluda, very good notes! You are absolutely right about name (prefix "with" is much better). Also note about "immutability" is fitting. I'll look at that more to think about it if it is possible to change it.
Thanks!

tonysneed:
Alright, I am finally back on my computer -- it's so much nicer than reading messages on a phone. :dizzy_face: 

Let me just reply to your [question](https://github.com/TrackableEntities/trackable-entities/pull/114#issuecomment-172852318):

> Do we only allow with interceptor (any, in the future, not only this current StateChangeInterceptor) to change entity state?

Yes, anyone can implement `IStateInterceptor` by implementing the `GetEntityState` method any way they want.  In `StateChangeInterceptor` we are a delegate of type `Func<TEntity, RelationshipType, EntityState?>`, but a future implementation of `IStateInterceptor` might do it differently.  For example:

```csharp
// Implements IStateInterceptor
public EntityState? GetEntityState(ITrackable item, RelationshipType relationshipType)
{
    if (item == null) return null;
    var unchangeable = item as IUnchangeable;  // Some interface
    if (unchangeable != null && unchangeable.IsUnchangeable)
        return EntityState.Unchanged;  // tell `ApplyChanges` to mark entity as Unchanged
    return null;  // Allow `ApplyChanges` to set state normally
}
```

There is no limit on the how the implementer of `IStateInterceptor` might want to set entity state.  He could, for example, look up entities in a cache and make the decision based on that.

```csharp
// Implements IStateInterceptor
public EntityState? GetEntityState(ITrackable item, RelationshipType relationshipType)
{
    if (item == null) return null;
    var customer = item as Customer;  // Some entity type
    if (customer != null && _cache.Contains(customer))
        return EntityState.Added;  // tell `ApplyChanges` to mark entity as Added
    return null;  // Allow `ApplyChanges` to set state normally
}
```

I think I see what you may be getting at with your `ShouldApplyChangesOnProperties` example, which would allow the caller to tell `ApplyChanges` whether or not to set state on an entity.  But I am not sure that is the best example, because the purpose of this extensibility point would be to allow a caller to tell `ApplyChanges` a) what state the entity should be set to, or b) allow `ApplyChanges` to set the state as it normally would (by returning `null` from `GetEntityState`.  I would not want to allow a caller to tell `ApplyChanges` to not set the state unless the caller specifies what that state should be.  So maybe you have another example in mind which could not be done by implementing `IStateInterceptor.GetEntityState`?

tonysneed:
Thanks @azabluda for your [comment](https://github.com/TrackableEntities/trackable-entities/pull/114#issuecomment-172828520) proposing a discussion of immutability, which is definitely worth pursuing.

If there's no objection what I would like to do at this point would be to **rename** all the methods in `InterceptorExtensions` from `AddXxxInterceptor` to `WithXxxInterceptor`.  Then hopefully @luboshl can perform the changes I proposed in my review of his PR. :pray:  Then I can accept and merge his PR, and we can open another PR to address immutability.

I prefer to have multiple PR's that are focused on one issue at a time, rather than one *monster* PR. :fearful:

azabluda:
Implementation of immutability is rather straightforward. I could do it once this PR is merged.

tonysneed:
Terrific, I'll merge #114 as soon as Lubos makes the changes I gave in my PR review, with the rename you proposed.

Then you can create a PR for your immutability feature. Thanks in advance!

luboshl:
So, I made all (resp. almost) changes based on these comments. To some of them I answered, so they are opened to change.

But - you wrote in [previsous comment] (https://github.com/TrackableEntities/trackable-entities/pull/114#issuecomment-172899681)
>If there's no objection...

And I have one :smile: :smile: You probably still don't understand, what I mean in my [comment] (https://github.com/TrackableEntities/trackable-entities/pull/114#issuecomment-172834269) (and some previous). Probably it is due to my bad English :confused: Still I made commit with changes based by comments in this PR so we can move on. Then @azabluda can make proposed changes (refactore fluent). During this time we can continue in discussion, I will try to explain it again.

luboshl:
So, I try to explain again our two understandings of interceptors.

#### Option 1

`IStateInterceptor` is interface of interceptors, that allows to **change the state of an entity** explicitly.
`StateChangeInterceptor` is concrete implementation, that is able to change the state by callback defined with `Func<TEntity, RelationshipType, EntityState?>`.
Then we can have any other implementation, the one from [your exmaple] (https://github.com/TrackableEntities/trackable-entities/pull/114#issuecomment-172892550)
with `IUnchangeable` or some with `IAllowedStates`, which ensures, that entity will stay in one of allowed states.

The main point is, that all of them are used to **change the state of an entity**.

#### Option 2

`IInterceptor` is inteface of interceptors, that allows to **change process in `AcceptChanges` in general**.
In [my example] (https://github.com/TrackableEntities/trackable-entities/pull/114#issuecomment-172834269)
`ShouldApplyChangesOnPropertiesInterceptor` is interceptor, that **doesn't change the state of an entity**. It does something else.
That one indicates whether to go through all descendants of given entity or not. This made-up interceptor has method:

```C#
internal bool ShouldApplyChangesOnProperties(ITrackable item)
{
    var entity = item as TEntity;
    return entity != null ? this.predicate(entity) : true;
}
```

And concrete usage:

```C#
context
    .WithShouldApplyChangesOnPropertiesInterceptor<Order>(e =>
    {
        // process descendants if order ID == 1
        if (e.OrderId == 1)
            return true;

        // do not process descendants if order ID == 2
        if (e.OrderId == 2)
            return false;

        // default, descendats are processed
        return true;
    })
    .ApplyChanges(order);
```

No state has been changed, only descendant for Order ID 2 are not processed in `ApplyChanges`. So that interceptor cannot implement `IStateInterceptor`,
because has no `GetEntityState` method. So in this option `IInterceptor` is simple, withou any method (I have no idea about something common). There could also exist
`IStateInterceptor` but as concrete interface implementing `IInterceptor`. For `ShouldApplyChangesOnPropertiesInterceptor` there would be something like
`IShouldApplyChangesOnPropertiesInterceptor`, but this is realy fake sample :smile: But that way I try to explain why `IInterceptor` was without any method.

In both options all interceptors are passed to `ApplyChanges` method. If it is the interceptor that changes the state of an entity, then `GetEntityState` is called.
If it returns *null*, continue normally. If returns some value, set state of entity to that state. If it is the interceptor that decides whether to process all descendants or not,
then its method `ShouldApplyChangesOnProperties` is called. That probably affects call of `context.ApplyChangesOnProperties` method in `ApplyChanges`.

That way I try to defend common `IInterceptor`. So my final proposal is:

Interfaces:
```C#
public interface IInterceptor
{
}

public interface IStateInterceptor : IInterceptor
{
    EntityState? GetEntityState(ITrackable item, RelationshipType relationshipType);
}
```

`IList<IStateInterceptor>' changed to `IList<IInterceptor>` in `ApplyChanges` method and in `InterceptorPool` class:

```C#
public static class DbContextExtensions
{
  internal static void ApplyChanges(this DbContext context, ITrackable item, IList<IInterceptor> interceptors)
  {
  }

  internal static void ApplyChanges(this DbContext context, IEnumerable<ITrackable> items, IList<IInterceptor> interceptors)
  {
  }
  
  // other methods
  ...
}

public class InterceptorPool
{
    internal DbContext DbContext { get; private set; }
    internal IList<IInterceptor> Interceptors { get; private set; }

    internal InterceptorPool(DbContext dbContext)
    {
        this.DbContext = dbContext;
        this.Interceptors = new List<IInterceptor>();
    }
```

My point is to have signature of `ApplyChanges` with `IList<IInterceptor>` where `IInterceptor` is interface for common interceptor, not for "concrete"
interceptor for change the state of entity.

I hope it is clear now. If so, we can decide now, whether to make `IStateInterceptor` **with** `IInterceptor` and to be prepare to the future interceptors,
or we don't want to and make only `IStateInterceptor` for changing state.

Good night :yum:

luboshl:
For the future, is it possible to debate also somewhere else here in GitHub? Not Issue, not Pull Request...

tonysneed:
AFAIK, For public discussions GitHub only has these two.  Issues are for any kind of discussion, including questions, debates, bug reports, etc, but these are without associated code, except for short snippets in markdown.

But when you want to show so code in context and discuss it, then a Pull Request is better than an issue.  In fact, for your proposals I think you could create another experimental branch and write code which demonstrates your design proposal. Then we can use line comments to discuss it. IMO this is so much better than code snippets in a discussion, because we can also see compiler errors and warnings, as well as refactoring suggestions from tools like ReSharper (BTW I can get you a free license!).

For private discussions we can use Gitter to create a private room, and of course we can chat there in real-time. :)

BTW, I am a believer in lively discussions.  I always learn something, and it's much more fun than coding alone. ;)

Is there something else you are looking for?

tonysneed:
So I carefully read your [comment](https://github.com/TrackableEntities/trackable-entities/pull/114#issuecomment-173016120), and I think I understand better now.  BTW, this is the kind of proposal that I think would be better served if you were to create a *temporary* **pull request**, in which you include your code snippets.  This would allow me to comment on specific lines of code in context and to see any compiler warnings or refactoring suggestions from ReSharper (R# also applies style guidelines).  So keep in mind for the future that you could always create a local branch for your proposal, publish it to your fork, then submit a PR.  Issues and PR's both allow for discussions, but PR's allow you to discuss proposals in context -- which is why I like them no much. :grinning: 

It seems like the gist of your proposal is to allow for greater extensibility by inheriting `IStateInterceptor` from `IInterceptor` but has a `GetEntityState method.  `IInterceptor` would have no methods, but someone else could then create an `INonStateInterceptor` that has a method such as `ProcessDescendents`, which returns `bool` instead of `EntityState?`.  This would allow some interceptors to directly affect entity state, while other interceptors could do something else, such as instruct `ApplyChanges` not to process any descendants.

So there are two aspects to your proposal.  One is the *idea* of a general interception mechanism, where the interceptor can do more more than suggest to `ApplyChanges` how to set the entity state.  While I like the idea in general, I'm not so sure I like it with regard the purpose of `ApplyChanges`, which is to paint entity state on an object graph.  What I'd like to do with this feature is to stay focused on entity state, rather than allow external control of, for example, how the graph is processed.  It's not that we couldn't consider this extensibility model in the future, but I would first want to see if there might be a compelling *use case*. With regard to interception for the purpose of dictating to `ApplyChanges` how to set entity state, I think we do have a compelling use case, which is the M-M scenario, where the caller should be able to indicate to `ApplyChanges` whether an entity should be created or merely added to the relation, based either on metadata, such as an extended tracking state property (`ITrackingStateExtended') or some domain-specific rules (for example, an identity property with a value of zero).

The second aspect of your proposal pertains to *implementation*: using what is commonly referred to as a *marker interface*, that is, an interface which does not have any members, from which other interfaces may inherit in order to provide different contracts.  I am not a big fan of marker interfaces and prefer to avoid them.  IMO they can make for messy code, as described in this SO [answer](http://stackoverflow.com/a/12688851), as well as the MS [design guidelines](http://stackoverflow.com/a/3310950).  If we were to decide to allow for non-state extensibility (which I'm not convinced we should do at this point), then we would need to look at a different design pattern to achieve this goal.

So for now I think the best way to proceed would be to just have `IStateInterceptor` with a `GetEntityState` method, so that callers can indicate to `ApplyChanges` how to set entity state based on whatever algorithm is specified by the caller (the Strategy design pattern), along with a fluent API.  Let's go with that for now, and I think we will have a good solution for the M-M problem, as well as an extensibility hook for `ApplyChanges` that will provide enough external control for most common use cases.

tonysneed:
I reviewed  commit 6f9554311dc494d50c9ce92d1f5da0debdbe2a35, and I think we should hopefully be able to merge your PR after the other changes are made.  If you need to discuss the issue of the `internal` methods, please let me know.  We can chat in real-time if necessary.

luboshl:
Many comments, so I will answer them one by one.

The most important is [your commment] (https://github.com/TrackableEntities/trackable-entities/pull/114#issuecomment-173094880) about whether to use common `IInterceptor` for any interceptor or specific `IStateInterceptor`. I am glad that you understand now what I ment :smile:

>While I like the idea in general, I'm not so sure I like it with regard the purpose of ApplyChanges, which is to paint entity state on an object graph. What I'd like to do with this feature is to stay focused on entity state, rather than allow external control of, for example, how the graph is processed.

... and...

>So for now I think the best way to proceed would be to just have IStateInterceptor with a GetEntityState method, so that callers can indicate to ApplyChanges how to set entity state based on whatever algorithm is specified by the caller (the Strategy design pattern with a fluent API). Let's go with that for now, and I think we will have a good solution for the M-M problem, as well as an extensibility hook for ApplyChanges that will provide enough external control for most common use cases.

I absolutely agree. I only wanted to be satisfied that you knew what I ment and we can do the best decision now. So, this is the current state, nothing to change in this case (interfaces etc.) :+1:

luboshl:
I hope I haven't missed any your comment. I made new [commit] (https://github.com/luboshl/trackable-entities/commit/2403db4efb8640aede8e43299cc0173e49c7d2c8).

You mentioned ReSharper few times. I know this tool, I am used to use it every day.

>..we can also see compiler errors and warnings, as well as refactoring suggestions from tools like ReSharper (BTW I can get you a free license!).

Because I have licence only from my employer and I cannot use it for my private purpose, it will be great, if you can give me a free licence! :pray:

I missed R# here (you know, `this.`, `_selector`...). I had in my mind to write an Issue to add team R# settings for the solution (it means `TrackableEntities.sln.DotSettings`) to be able to see green marks everywhere :smile:

tonysneed:
Thanks for implementing my suggestions and pushing a new commit.  I'll have a look at that.  Contact me privately for the R# key for OSS dev.  My email is: tony@tonysneed.com.  I have heavy spam filters on that email, so I don't mind making it public here. :smile:

luboshl:
I sent an email so I hope it wasn't cathed by spam filter :smile:

tonysneed:
I don't have any other changes for your last commit to this PR -- we're looking really good now. :smile: 

Now it's time to implement `DbContextExtensions.ApplyChanges`, so we can get our tests to pass.

First we'll need to add a private `GetRelationshipType` method that returns `RelationshipType`.  See the `IsRelatedProperty` method.

Then you can add an optional `pool` parameter to `ApplyChanges`:

```csharp
private static void ApplyChanges(this DbContext context,
    ITrackable item, ITrackable parent, ObjectVisitationHelper visitationHelper,
    string propertyName, TrackingState? state = null, InterceptorPool pool = null)
```

On line 204, you can get the relation type, pass it to the pool of inspectors.

```csharp
if (pool != null && pool.Interceptors != null)
{
    if (parent != null && propertyName != null)
    {
        var relationType = GetRelationshipType(context, parent.GetType(), propertyName);
        foreach (IStateInterceptor interceptor in pool.Interceptors)
        {
            var entityState = interceptor.GetEntityState(item, relationType);
            if (entityState != null)
            {
                context.Entry(item).State = (EntityState)entityState;
            }
        }
    }
}
else
{
    context.Entry(item).State = item.TrackingState.ToEntityState();
}
```

This is just a rough sketch, so feel free to clean it up as needed.

luboshl:
I messed it up :worried: I made squash, but badly, obviously. The more I tried to repir it the worse result it was :worried: Then I deleted the branch. But commits remained in this PR. The PR was closed automatically.

Crated new PR #117.

luboshl:
I'll never do `squash` until I am more familiar with git, I promise :blush:

tonysneed:
@luboshl, I understand .. I've had problems with interactive rebase as well.  What I would suggest is that, in the future, you don't try to do a squash commit until the very end.  Just before I'm about to merge, what I'll do is *ask* you at that point to create a squash commit.  There's no need to squash before that point, because there will likely be more commits that you'd have to squash.  So better to wait until I give you the green light and am ready to merge.  Then at that point you can try doing a squash commit -- after first making a backup. :smile:

azabluda:
Yeah, and "making a backup" in Git is as easy as tagging the top revision before you start rebasing. Thus even if you mess things up completely, you can always reset --hard to that tag, rethink and restart.

luboshl:
I am newbie to Git (only one week), so many commands and their parameters :smile: As I wrote and also @tonysneed - fot the next time I will wait before I do reabase. But I'll have a look at *reset --hard* and tagging.

I did rebase with squash before, there were some conflict. I merged them and everything was fine. I do not know, what happend now. Commits are duplicated, disarranged... Sorry for that mess.

azabluda:
No worries mate! Git is known for being difficult to learn yet extremely powerful. I found [TortoiseGit](https://tortoisegit.org/) quite helpful for visualization of branches. Prior to pushing a PR I used to use it for checking whether my branch contains only relevant commits as well as commits contain only relevant changes (e.g. _TrackableEntities.sln.DotSettings_ file doesn't seem to belong to #117, etc.)

Of course you may try any other [graphical Git tool](https://git-scm.com/downloads/guis).

tonysneed:
I like TortoiseGit and use it quite a bit.  I also prefer to leverage the Team Explorer window in Visual Studio 2015, especially with the GitHub piece, which you can get here: https://visualstudio.github.com <https://visualstudio.github.com/>.  Those take care of 90% of my needs.  However, there are times when the command line is needed, for example, for interactive rebase and squatting commits.  I also use the command line when reviewing pull requests, because I can first pull the commits from the fork into a separate branch, then review the changes.

luboshl:
Thak you guys. I uninstalled TortoiseGit yesterday, because I haven't used it. I failed to notice there is [Revision Graph] (https://tortoisegit.org/docs/tortoisegit/images/RevisionGraph.png). I have read about it mainly as Explorer integratin tool. And I don't use Window File Explorer if I don't have to :smile: But the Revesion Graph looks awesome!

Git Extension in Team Explorer in VS 2015 I'am already using, Git for Windows as well. But I would like to learn git commands, because I can see they are much powerful (but also dangerous). What a shame that graphical tools dont show what git command is used when user select something.

tonysneed:
> What a shame that graphical tools dont show what git command is used when user select something.

You can see the Git commands by opening the Output window and selecting Git from the drop down.

Have you installed the VS 2015 **GitHub** integration?

luboshl:
Realy? Great, I'll watch the Output.
And yes, I've got exactly "VS 2015 GitHub" intergration, not just common Git integration :+1:

luboshl:
I checked it and I cannot see git commands exactly (in the Output window in VS). I can see some log info like "Commit 99ac8489 created locally in repository ..." but not "git commit".

tonysneed:
Yes, I think it might not show all the Git commands.  I was thinking of “Visual Studio Code,” which does show all the Git commands.

tonysneed:
This is a bit outdated by now, but it helped me to learn how to use the GUI in VS for Git: https://www.visualstudio.com/en-us/get-started/code/share-your-code-in-git-vs <https://www.visualstudio.com/en-us/get-started/code/share-your-code-in-git-vs>.

PR #117, opened 2016-01-22
Implementation of state interceptors.

luboshl:
Full implementation of interceptors for change the state of entities. This PR replaces previsous PR #114.


luboshl:
All tests passed! But we need some more, for different relationships for example.

luboshl:
Some more notes to https://github.com/TrackableEntities/trackable-entities/pull/114#issuecomment-173454429 about the implementation:

* I changed signatrue of `ApplyChanges` - we do not need `InterceptorPool` as argument, just collection of interceptors. 

* I call `ChangeStateByInterceptors` at many places in `ApplyChanges`, not only on line 204. Becasue we need to cal interceptors in all cases where the state of entity is changed.

* I didn't know if it is better to call interceptors **before** the state of entity is normally changed in `ApplyChanges` or **after** that. I mean this:

After (that's how I implemented that):
```C#
context.Entry(item).State = item.TrackingState.ToEntityState();
ChangeStateByInterceptors(context, item, parent, propertyName, interceptors);
```

Before (would be somethnig like this):
```C#
bool wasInterceptorApplied = ChangeStateByInterceptors(context, item, parent, propertyName, interceptors);

if (!wasInterceptorApplied)
  context.Entry(item).State = item.TrackingState.ToEntityState();
```

For now I have done it the easier way, that is the *after* option. The interceptor can use the current state of the entity after normal process changes it based on common flow.

So what do you think?

tonysneed:
In response to your [comment](https://github.com/TrackableEntities/trackable-entities/pull/117#issuecomment-173871230) above:

> I changed signatrue of ApplyChanges - we do not need InterceptorPool as argument, just collection of interceptors.

This is fine.

> I didn't know if it is better to call interceptors before the state of entity is normally changed in ApplyChanges or after that.

We should change `ChangeStateByInterceptors` to `SetEntityState` and refactor as shown [here](https://github.com/luboshl/trackable-entities/commit/99ac84896ca29c029a2c9a6ec8734847e34f125f#commitcomment-15628385).  This will encapsulate the logic of when to invoke the interceptors and when not to.  If interception is possible, then we will use it, otherwise we'll set the entity state normally.

> I call ChangeStateByInterceptors at many places in ApplyChanges, not only on line 204. Because we need to cal interceptors in all cases where the state of entity is changed.

The rule is that we call the new `SetEntityState` method wherever there is this line of code:

```csharp
context.Entry(item).State = state
```

Then we just need to pass an `IList<IStateInterceptor>` wherever we need to call `SetEntityState`, which will include not only `ApplyChanges`, but also supporting methods such as `ApplyChangesOnCollectionProperties`, `SetChanges`, etc.

luboshl:
First of all we need to clarify using of (non) nullable `RelationshipType`, because it is important and many your comments depend on it.

Argument `relationshipType` in `IStateInterceptor.GetEntityState(...)` realy needs to be nullable. If I have entity without any relationship, this argument must be null. But it's true that `DbContextExtensions.GetRelationshipType(...)` can be changed to retun non-nullable `RelationshipType`. When this method is called, there is *if* condition:

```C#
RelationshipType? relationType = null;

if (parent != null && propertyName != null)
    relationType = GetRelationshipType(context, parent.GetType(), propertyName);
```

Then `GetRelationshipType` realy can throw an exception if no proper relationship is found. As you can see in the code snippet above, `GetRelationshipType` is called only for entities with a relationship. So we still need nullable `RelationshipType` in `IStateInterceptor.GetEntityState(...)`.

tonysneed:
Yes, it's first address the issue of nullable vs non-nullable for `RelationshipType`.  Notice in my [comment](https://github.com/TrackableEntities/trackable-entities/pull/117#commitcomment-15628385) the following condition:

```csharp
// Set state normally if we cannot perform interception
if (interceptors == null && parent == null && propertyName == null)
{
    context.Entry(item).State = state;
    return;
}
```

So we are first checking to see if it is possible to get the relation type, and we only call `GetRelationshipType` if `parent` is not `null`.  So there is absolutely no need to use `RelationshipType?`.  I would prefer to throw an exception instead of returning `null` from `GetRelationshipType`.  If we cannot get the relationship type, then we throw. Thanks!

luboshl:
I agree that `GetRelationshipType` should return null. There is no problem.

But in your [comment] (https://github.com/TrackableEntities/trackable-entities/pull/117#commitcomment-15628385) after 

```C#
// Set state normally if we cannot perform interception
if (interceptors == null && parent == null && propertyName == null)
{
    context.Entry(item).State = state;
    return;
}
```

there is

```C#
// If there are any state interceptors, call them to try to get state
RelationshipType relationType = GetRelationshipType(context, parent.GetType(), propertyName);
foreach (IStateInterceptor interceptor in interceptors)
{
    // If current interceptor returns the state, use it
    var entityState = interceptor.GetEntityState(item, relationType);
    if (entityState != null)
        context.Entry(item).State = (EntityState)entityState;
}
```

This code will be called some interceptors (`interceptors != null`) because the condition above is not met. Then the method would throw and exception (as we agreed) if `parent == null || propertyName == null). So it must be changed to

```C#
RelationshipType? relationType = null;

if (parent != null && propertyName != null)
    relationType = GetRelationshipType(context, parent.GetType(), propertyName);

foreach (IStateInterceptor interceptor in interceptors)
{
    // If current interceptor returns the state, use it
    var entityState = interceptor.GetEntityState(item, relationType);
    if (entityState != null)
        context.Entry(item).State = (EntityState)entityState;
}
```

That is the reason why `GetEntityState` must return nullable `RelationshipType`.

tonysneed:
> This code will be called some interceptors (interceptors != null) because the condition above is not met. 

This should not be the case because `||` will be short-circuited?

luboshl:
I don't understand. If we have some interceptors (it means `interceptors` is not null), this will be skipped:
```C#
// Set state normally if we cannot perform interception
if (interceptors == null && parent == null && propertyName == null)
{
    context.Entry(item).State = state;
    return;
}
```

Next code is this:
```C#
// If there are any state interceptors, call them to try to get state
RelationshipType relationType = GetRelationshipType(context, parent.GetType(), propertyName);
foreach (IStateInterceptor interceptor in interceptors)
{
    // If current interceptor returns the state, use it
    var entityState = interceptor.GetEntityState(item, relationType);
    if (entityState != null)
        context.Entry(item).State = (EntityState)entityState;
}
```

And there, `parent` is null, `propertyName` is null. So frist, it whill throw `NotNullException` at `parent.GetType()` statement. I cannot call method `GetRelationshipType` here.

Concrete sample:
```C#
var context = TestsHelper.CreateNorthwindDbContext(CreateNorthwindDbOptions);
var product = new Product();
product.TrackingState = TrackingState.Added;

context
    .WithStateChangeInterceptor<Product>((e, rs) => EntityState.Unchanged)
    .ApplyChanges(product);
```

tonysneed:
Regarding this [comment] of yours, what I'm thinking is that if order to save time, why don't you try implementing the refactorings I've proposed and put them in a separate commit.  We can then then discuss the pros and cons more concretely, by commenting on specific lines of code in that commit.

One of the great things about Git is that you can easily reset to a prior commit if things don't work out.

I'm also thinking that if you think something is incorrect or can't work, then demonstrate it with a failing test.  That should prove the case you are making. :smile:   I think it should be much more productive and less time-consuming to go about it in this way.  What do you think?

luboshl:
>what I'm thinking is that if order to save time, why don't you try implementing the refactorings I've proposed and put them in a separate commit.

Ok, no problem. But just before it please see [this comment] carefully

luboshl:
Sorry, wrong button, I didn't mean to close this... Can you un-close it?

tonysneed:
oh yes ..

tonysneed:
Responding to his [comment](https://github.com/TrackableEntities/trackable-entities/pull/117#issuecomment-174168794) .. You're right that we need to be able to call an interceptor without supplying the `RelationshipType`, because we want to allow interception at the top level of an object graph.

We also need to refactor `SetEntityState` so that entity state is set when the `parent` is null. So please proceed on that assumption.  Thanks!

tonysneed:
I would if possible like to return a non-nullable `RelationshipType` from `GetRelationshipType`.  But we can make it nullable for `IStateInterceptor.GetEntityState`.

luboshl:
Great, that's it! That corresponds to my comments. Now I have lunch time, then I prepare some changes and commit them.

tonysneed:
Yay! 😜

tonysneed:
I think you still to do [this](https://github.com/TrackableEntities/trackable-entities/pull/117#commitcomment-15630602) - replace `TrySetEntityState` on line 134 with a // TODO

luboshl:
Aha, misundersanding. I thought it was about https://github.com/luboshl/trackable-entities/commit/d7850127c4eb10a3e986afd30a752b92250d9b14. Ok, replaced with TODO and previous returned back.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -624,12 +657,82 @@ private static EntityState ToEntityState(this TrackingState state)
...
+            if (navProp.FromEndMember.RelationshipMultiplicity == RelationshipMultiplicity.Many
+                && navProp.ToEndMember.RelationshipMultiplicity == RelationshipMultiplicity.Many)
+                return RelationshipType.ManyToMany;
+
+            if ((navProp.FromEndMember.RelationshipMultiplicity == RelationshipMultiplicity.One
+                 || navProp.FromEndMember.RelationshipMultiplicity == RelationshipMultiplicity.ZeroOrOne)
+                && navProp.ToEndMember.RelationshipMultiplicity == RelationshipMultiplicity.Many)
+                return RelationshipType.OneToMany;
+
+            throw new InvalidOperationException(String.Format("Cannot determine relationship type for {0} property on {1}.", propertyName, entityType.FullName));
```
tonysneed: Change `String` to `string` - it will make R# happier. :smile:

tonysneed:
The only thing left at this point (besides [this](https://github.com/TrackableEntities/trackable-entities/pull/117/files#r50620871)) is to add some more tests.  We'll need tests for various relationship types: 1-1, 1-M, M-1, M-M.  We'll want to use `RelationshipType` in some of these.  We'll also want to test multiple interceptors, both for a single type as well as multiple types.  We also may want a test or two for some other implementation of `IStateInterceptor`.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
luboshl: R# will be happy, but I will be sad :wink: Just kidding. When I call static methods of `String` object, I preffer to use `String` instead of `string`, becouse it really is a class. Opposite to `string str = ""` is better than `String str = ""`. But ok, I'll change it. BTW I am preparing full R# defaults. I'll write about it later and not in this PR...

luboshl:
What do you mean with
>The only thing left at this point (besides [this] (https://github.com/TrackableEntities/trackable-entities/pull/117/files#r50620871)) 

Is there something I missed?

tonysneed:
Yes, just the `string` thing was missing at that point.

luboshl:
Ok :smile:

tonysneed:
Can you also remove `Source/TrackableEntities.sln.DotSettings` from this PR?  Those you can submit in a separate PR -- but I imagine that the R# settings file can just contain the default settings?

luboshl:
Done :smile:

luboshl:
I will prepare one test for our TODO (https://github.com/luboshl/trackable-entities/commit/a33269b3362fa000287463ed958a9137dace54f4)

tonysneed:
Please put the test for the TODO in a separate PR. Thanks!

luboshl:
But it is related to interceptors. I cannot make failing test for branch `develop` where is no interceptors. What branch shall I make? Some new branch from `develop`? Or am I misundersdaning?

tonysneed:
You can create a new branch off of iss106 and add it there.  But what I suggest is to first write the other tests which will be needed in order for me to merge this PR.  Then you can create a new branch off develop for the TODO PR.  Does that make sense? :smile:

luboshl:
Yes, it now makes sense :smile:

luboshl:
Maybe I found an issue on line 147. See this case:
- call SetEntityState on line 147
- call TrySetEntityState on line 734
- call GetRelationshipType on line 705
- method GetRelationshipType throws the `InvalidOperationException` ("Cannot determine relationship...") on line 690
- exception `InvalidOperationException` is catched on line 149

But there was excpected other `InvalidOperationException` before we added other to line 690. Then rethrown exception`throw new InvalidOperationException(Constants.ExceptionMessages.DeletedWithAddedChildren, invalidOpEx);` doesn't make sanse.

luboshl:
We could add other failing test for this `InvalidOperationException` issue. I found existing test `Apply_Changes_Should_Mark_Unchanged_Order_Deleted_Customer_With_Addresses_Multiple_Added` which tests the original exception. I'll prepare other for [previsous case] (https://github.com/TrackableEntities/trackable-entities/pull/117#issuecomment-174193537).

tonysneed:
For [this](https://github.com/TrackableEntities/trackable-entities/pull/117#issuecomment-174193537) we could create a custom exception and throw it from `GetRelationshipType`, rather than `InvalidOperationException`.  Strings for exception message should reside in `Constants.ExceptionMessages`.

luboshl:
I've created test for [this] (https://github.com/TrackableEntities/trackable-entities/pull/117#issuecomment-174190893). It is realy not needed there. Child entity is set to `Deleted` in `SetChanges` so `SetEntityState` is called to apply interceptors.

Check [this commit] (https://github.com/luboshl/trackable-entities/commit/6747401843b6ec2dfdb3127cbeb33154a59f44a6)

tonysneed:
Looks good so far.

luboshl:
I think we only need some more test. You wrote:

>We'll need tests for various relationship types: 1-1, 1-M, M-1, M-M. We'll want to use RelationshipType in some of these. We'll also want to test multiple interceptors, both for a single type as well as multiple types. We also may want a test or two for some other implementation of IStateInterceptor.

Do you have some concrete cases or combinations?

tonysneed:
I am thinking we could use a [Theory] xunit attribute to run a single test multiple times with different parameters.  We migrated our tests from NUnit to xUnit, but we never really took advantage of this ability in xUnit and might as well use it now.

So you could start with some mock orders, from which you should be able to do M-1 (Order-Customer), 1-1 (Customer-CustomerSettings), 1-M (Order-OrderDetail).  For M-M you would use Employees-Territories.  You can see example in other tests we've written.  So you could add interceptors for each of these.  There could be a set of tests which use just one interceptor.  Then another which uses multiple interceptors, for example for different relations in the graph.

This should be enough to get you started ... :smile:

luboshl:
Ok, so first to try `[Threry]` I refactored four test method to use two methods with `[Theory]` insted. It is fine that I need less methods and I can only add more inputs without creating other method.

But it looks little unclear for the first look.

So I changed these methods

```
Apply_Changes_With_State_Selector_Should_Mark_Multiple_Entities_Added_Modified_Deleted
Apply_Changes_With_State_Selector_Should_Mark_Multiple_Entities_Unchanged
```
to
```
Apply_Changes_With_State_Selector_Should_Change_Multiple_Entities_State
```

And methods
```
Apply_Changes_With_State_Selector_Should_Mark_One_To_Many_Entities_Added_Modified
Apply_Changes_With_State_Selector_Should_Mark_One_To_Many_Entities_Unchanged
```
to
```
Apply_Changes_With_State_Selector_Should_Change_One_To_Many_Entities_State
```

If you agree with [that form] (https://github.com/luboshl/trackable-entities/commit/a3479f71c3e2f854eca19fc398086c8016104396), I'll then create the new methods.

luboshl:
BTW in this way all older test methods like `Apply_Changes_Should_Mark_Product_Unchanged` and `Apply_Changes_Should_Mark_Product_Added` etc. can be merged to one with `[Theory]` very easy :smile:

tonysneed:
Looking good - yes please continue! :))

tonysneed:
I reviewed and commented on your commits, *except* the last two (21619497263ca2fddc678ff9e0b1f727040aa7b2 and 297637a643e5599e39864b834edc9f3a006ccd7e), which I'll have a closer look at today.  Thanks!

luboshl:
`this.` again... I'll have to check R# settings to help me with that :smile:

tonysneed:
Here is what R# is showing me.

rs-this

luboshl:
I think it's not R# but PostSharp. I can see it too :smile: But I have green symbol (in scrollbar) that everythink is ok.

tonysneed:
Do you see the lightbulb in the margin and is `this.` a lighter shade?

tonysneed:
I can see why you like [PostSharp](https://visualstudiomagazine.com/articles/2012/09/01/aop-with-postsharp.aspx) -- the company is headquartered in Prague. :smile:

luboshl:
I can see this:

image

Only `this.` is a little bit lighter shade, as you wrote. But as I am used to use `this.`, I cann se the little difference :)

tonysneed:
Obviously PostSharp doesn't care about certain things.  What happens when you put your cursor on line 28 and press Alt + Enter?  Do you see the lightbulb and list of remedies?

tonysneed:
Alright, I think I have some suggestions on how to structure the tests and set up the code regions.

*Great job BTW on usage of the* `[ClassData]` *attribute with* `[Theory]` *tests.* :clap: 

First of all, it looks like the **NorthwindDbContextTests.cs** file is getting quite large, and it's not really very descriptive.  We should add a file called **ApplyChangesStateInterceptorTests.cs** and move all of our interceptor tests there.  Then we should create the following regions for groups of tests:

- No Relationship
- 1-1 Relationship
- 1-M Relationship
- M-1 Relationship
- M-M Relationship

Then within each group, we should have a systematic approach, to make sure we've covered all our bases.  So for example, we could have a set of tests with a single top-level interceptor, then another set of tests with top-level and second-level interceptors.  Within each of these groups, we could then have tests for setting state without regard to the relationship type, and also for setting it based on relationship type.  We'll need to test for certain state transitions, for example, Added to Unchanged, or Unchanged to Modified (make sure also to set ModifiedProperties).  But I don't think we'll need to cover all possible state changes -- we just want to make sure we cover enough to make us feel good about the code quality.  We'll also want to set for expected exceptions, and xUnit has a special Assert for that.

*This is just a suggestion, so no need to write tests for all the possibilities.  I realize that writing the tests can be time consuming, so we just want to cover the main scenarios.*

Then we'll want to add a file called **ApplyChangesInterceptorTests.cs** which uses non-state change interceptors.  Again, we don't have to cover every possible permutation, but we can have some tests to document expected behavior and make sure we have some assurance on what the code is supposed to do.  Thanks!

luboshl:
You are right. Compare these screenshots:

With PostSharp:
image

Without PostSharp:
image

Redundant qualifier in static method call:
image

I think that PostSharp uses the same icon as VS and that it was realy only from VS. Compare to the third image - there you can see notification from R#, PostSharp and VS :smile:

luboshl:
To [your comment about tests] (https://github.com/TrackableEntities/trackable-entities/pull/117#issuecomment-175544805):

> *Great job BTW on usage of the* `[ClassData]` *attribute with* `[Theory]` *tests.* :clap:

Thanks :smile:

>First of all, it looks like the NorthwindDbContextTests.cs file is getting quite large, and it's not really very descriptive. We should add a file called ApplyChangesStateInterceptorTests.cs and move all of our interceptor tests there. Then we should create the following regions...

I totally agree!

And to the rest - ok :smile:

tonysneed:
Seems like it's hard to see the faded text with your color scheme.  I wonder if R# will let you change that?

tonysneed:
I'm looking forward to seeing the next round of tests! Rock on. :sunglasses:

luboshl:
I am sure I can change colors for R#. But I am afraid, this is not from R# and that it is not possible to change this `this.` settings. I would prefer to be able to set inspection warning, but I think this setting is away since last (or previsous) version of R# :worried:

tonysneed:
Try here ..

rs-code-inspect-settings

tonysneed:
Not sure if that's too helpful, because there are are lot of R#-specific options in Fonts and Colors, and it may not be apparent which one needs changing.

luboshl:
So, I've [added first tests for state interceptors] (https://github.com/luboshl/trackable-entities/commit/1dbbe1471f9e10aacbf56101ce5054bec22b5b41). For now all test data are automatically generated. It is not usable, because some test take very long time. But for now, we can focus on the failing tests. There are a lot of them, but probably it's because there are all possible combinations, but not all of them make sense.

Time of tests:
- 256 tests of 1-1 or "multiple entities" take less then 1 second
- 256 tests of M-M take approx. 8 seconds
- 256 tests of 1-M and M-1 takes approx. 1.5 minute

image

Do you think that in 1-M and M-1 tests it is so much more complex to go through the graph of entities, so it takes so much time?

What I don't understand are some failing tests. Probably I don't understand some cases. Here are some fo them:

1. M-M - Employee: Unchanged or Added, Territory: Added (no state interceptor) - why the final state of Territory is Unchanged and not Added? Why cannot I only add Territory?
2. M-M - Emplyee: Unchanged, Teritorry: changed by interceptor from Added to Deleted (or from Deleted to Added): Error "_Cannot change relationship's state to the state other than deleted or detached if the source or target entity is in the deleted state._" (resp. "_Cannot change relationship's state to the state other than added or detached if the source or target entity is in the added state._"). Why not? :smile:
3. 1-1 - Customer: Added, CustomerSetting: Unchanged (no state interceptor) - why the final state of CustomerSetting is Added and not Unchanged? What if CustomerSetting already exists in database, Customer is now created and the realtionship is created?

And so on... Please check results of the tests

tonysneed:
I'm wondering if we could put these tests in a separate branch and PR? This PR is already as big as it should be. Then we can break up the new PR into several commits. Maybe we can have an initial commit with just basic tests.

Today is a busy day with my family, but perhaps we can chat on Gitter about how best to proceed. I may need some help from Alex for this one. :)

luboshl:
Ok, no problem. A was also busy these days and today we plan to go on a trip with children. I will install Gitter to my mobile :smile:

tonysneed:
Cool - I didn't realize Gitter had a mobile app! I am also away from home, so I installed Gitter on my iPhone. I think I will be able pull your latest commits when I get back home later this afternoon.

luboshl:
(Only test to answer directly from mail)

tonysneed:
Actually I think we can handle these last two commits in this same PR #117. There aren't that many tests after all.  I was concerned about a large quantity of tests, but I think that's resulting from the amount of data.  I'll have a closer look at that.

luboshl:
Yes, there are only 6 test methods (plus one simple with `[Fact]') using `[Theory]` with `[ClassData]` which generate all the test data.

tonysneed:
I've looked at the first test: `State_Interceptor_Should_Mark_Single_Entity_State`.  The test passes, and here is the session results in the R# test runner:

state_interceptor_should_mark_single_entity_state

I only ran this one test and no others.  As you can see R# is reporting false failures, which we can ignore for now.  This one test is called **16 times** with different parameters.  So my general observation is that it may not be necessary to test for *each possible combination* of initial and final states.  I think we can be a bit more precise and save ourselves a lot of unnecessary test runs.

The best approach might be to look at where `ApplyChanges` calls  the interceptors to determine whether checking all the state combinations makes sense.  For example, from `State_Interceptor_Should_Mark_Single_Entity_State`, `ApplyChanges` calls the interceptors only from line 215, so various state transitions are all treated equally.

```csharp
// Set entity state
SetEntityState(context, item, parent, propertyName, item.TrackingState.ToEntityState(), interceptors);
```

In this case we want to test just **two conditions**.  One is where there is a state change *caused by the interceptor*, and the other is where the interceptor returns `null` and the state change *takes place normally*.  All the other test cases appear to be unnecessary.  Basically, we just want to be able to see if the interceptor is able to determine the final state, and we want to know if the interceptor returns `null` that `ApplyChanges` performs the default behavior.

So in this case, it seems like using `ClassData` may be self-defeating.  One problem is that it is not apparent from looking at the test what specific conditions we are actually testing for. Those are abstracted away in the `IEnumerable<object[]>`.  Second, the extra tests take more time to run and really don't add any value.  So we could refactor this first test as follows:

```csharp
[Theory]
// Interceptor should change state
[InlineData(TrackingState.Modified, EntityState.Unchanged, true)]
// Interceptor should not change state
[InlineData(TrackingState.Modified, EntityState.Modified, false)]
public void State_Interceptor_Should_Mark_Single_Entity_State(TrackingState initState, EntityState finalState, bool useInterceptor)
```

I'm not saying we should do away with `ClassData` in every case.  But we should determine if it is necessary for each one of the relationship types (including none).  Sorry I did not provide clearer guidance when you first asked about `ClassData`.  But let me know what you think about my feedback.  Thanks!

luboshl:
I don't think that `State_Interceptor_Should_Mark_Single_Entity_State` takes much time, in my environment it runned for 33ms. But as I wrote [in my previous comment] (https://github.com/TrackableEntities/trackable-entities/pull/117#issuecomment-177317891), "For now all test data are automatically generated". The important is "for now". I dont' expect we will have all tests with generated data. But maybe this first one with single entity could be a good candidate for that. On the other side...

>One problem is that it is not apparent from looking at the test what specific conditions we are actually testing for.

... it's true!

Next one, where also all combinations passed, is `State_Interceptor_Should_Mark_Multiple_Entities_States`. It is practically the same test, but with two entities, so that's why there is 256 combinations. It is completely unnecessary. I am not sure, if this test is necessary at all.

The `Last_Interceptor_Should_Win` is simple.

Remaining tests are more interesting. There are also all data generated from all possible combinations. It is surely also not necessary, it is **for now** only to determine what cases we should test. There are many fails and I don't understand some of them why they failed. It's up to you now to say, why they failed and whether it is ok they failed or not. And then to define scenarios which make sense and we should test. Don't hurry, I understand it consumes much time.

tonysneed:
I'll go thru the tests one at a time, and I'll propose changes.  Here is the [first one](https://github.com/luboshl/trackable-entities/commit/1dbbe1471f9e10aacbf56101ce5054bec22b5b41#commitcomment-15787219).

Eventually, I'll get to the ones that are failing and see what is happening there.

Regarding your [previous comment](https://github.com/TrackableEntities/trackable-entities/pull/117#issuecomment-177317891), the error you're seeing is from Entity Framework, which does not allow certain state transitions.  Here is a related [discussion](https://social.msdn.microsoft.com/Forums/en-US/35dadb23-548e-424a-8361-f0f1d2efbcbe/self-tracking-entities-applychanges-exception-when-deleting-entity-in-many-to-many-relation?forum=adodotnetentityframework) of this issue.

tonysneed:
@lubosh

tonysneed:
@luboshl, Just trying out a GitHub client for iOS. :)

Appreciate your work on this PR, and I think we're almost to a place where we can wrap it up.

I have an idea about how we might be able to proceed from here.  Are you able to chat on Gitter?

tonysneed:
This PR was merged into #118.

PR #120, opened 2016-02-10
Immutable interceptors

azabluda:
Hi @tonysneed and @luboshl,

As I briefly mentioned in one of our discussions, method `WithInterceptor` should not modify the upstream chain of interceptors. Please review the proposed solution.

### DIFF
--- a/Source/Tests/TrackableEntities.EF.5.Tests/StateChangeInterceptorTests.cs
+++ b/Source/Tests/TrackableEntities.EF.5.Tests/StateChangeInterceptorTests.cs
@@ -23,6 +23,25 @@ namespace TrackableEntities.EF5.Tests
 
         #region No Relationship (Product)
 
+        [Fact]
+        public void WithStateChangeInterceptor_Should_Not_Mutate_Pool()
+        {
+            // Arrange
+            var context = TestsHelper.CreateNorthwindDbContext(CreateNorthwindDbOptions);
+            var product = new Product();
+
+            bool calledX = false, calledY = false;
+            var poolX = context.WithStateChangeInterceptor<Product>((e, r) => { calledX = true; return null; });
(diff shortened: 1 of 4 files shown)


tonysneed:
Great PR .. I'll have a look!

tonysneed:
Thanks @azabluda, good catch!

luboshl:
Hey, @azabluda, it's nice, very nice! I like it!

tonysneed:
:+1:

luboshl:
@azabluda, do you know about a good guideline for fluent API? I've found
only some basic and simple.

azabluda:
Hey @luboshl, unfortunately I don't know any specific guidelines. But I think we just all intuitively know how fluent APIs are to look and to behave (type safe, immutable, etc). The particular implementation then would simply follow our expectations, and (as you can see) it doesn't have to be complex at all.

I'm glad that you liked my solution. :bow:

tonysneed:
This article by Martin Fowler may be helpful: http://martinfowler.com/bliki/FluentInterface.html.

One of the best examples of a fluent API is LINQ. Just use something like dotPeek to look at System.Linq.Enumerable.

I like Alex's PR to show the benefits of immutability, with the first commit containing just a failing test. I used Git checkout SHA to go to that commit and see the problem with a mutable InterceptorPool, which is that you were not able to get the same result from the first pool after calling the extension method again.

luboshl:
I know this article from Martin Fowler. It's the introdution to fluent API,
but I meant some recommendation about the implementation :-) For example
the immutability is something I didn't think about, when I implemented the
interceptors here in TE. So I am glad, that @azabluda mentioned that in a
comment earlier and for this PR :+1:

tonysneed:
Maybe one of us should write a blog post? :smile:

luboshl:
:-)

PR #134, opened 2016-08-05
Fix DBContext Inheritance

kourosko:


tonysneed:
Looks good - thanks so much!

PR #136, opened 2016-08-28
Added Many-to-Many test

MarcDrexler:
I think adding and removing the same territory to an employee should leave everything as unchanged.

The GetChanges method works as expected and returns 0 changes.
But the territory is still part of the employee's territory collection.
I have added a failing test.


tonysneed:
Merged this PR with #137, so that I can implement a fix in the new PR.

Issue #132, opened 2016-07-27
After removing an item that has Added state it goes to the _deleted internal collection with state Unchanged.


gtas-verisk:
So I have an object graph.

> MainObj -> TrackingCollection1<Obj1>
>                -> TrackingCollection2<Obj2>
>                -> TrackingCollection3<Obj3>

Say I create an instance Instance1 and has TrackingState = Unchanged, modify some properties, now TrackingState = Modified, then add it to TrackingCollection1<Obj1>.Add(Instance1), now TrackingState = Added, then modify a property, TrackingState = Modified again, now TrackingCollection1<Obj1>.Remove(Instance1) it is added to the internal collection with TrackingState = Deleted, all good.

Now in another scenario.  I create an instance Instance1 and has TrackingState = Unchanged, modify some properties, now TrackingState = Modified, then add it to TrackingCollection1<Obj1>.Add(Instance1), now TrackingState = Added, I do not modify again any properties and immediately remove it, TrackingCollection1<Obj1>.Remove(Instance1) it is added to the internal collection with TrackingState = Unchanged, this has an effect when calling GetChanges() that it takes the "Unchanged" item from the internal _deleted collection back to the TrackingCollection1<Obj1>, check the extension method below.

```
public static IEnumerable<T> GetDeletedEntities<T>(this ChangeTrackingCollection<T> list, IMainThreadDispatcherService dispatcherService) where T : class, ITrackable, INotifyPropertyChanged
        {
            IList<T> dtos = new List<T>();
            if (list.Any())
            {
                dispatcherService.RequestMainThreadAction(() =>
                {
                    dtos = list.GetChanges().Where(p => p.TrackingState == TrackingState.Deleted).ToList(); // Here the Unchanged item goes back to the ChangeTrackingCollection<T> list.
                });
            }
            return dtos;
        }
```

How to avoid this effect?

Update:

I see here, https://github.com/TrackableEntities/trackable-entities/blob/develop/Source/TrackableEntities.Client/ChangeTrackingCollection.cs#L256 that you have a try/finally code but you RemoveRestoredDeletes only in the finally block, but in the try block you return, I think the finally block is never invoked unless an exception, is this the desired functionality?

tonysneed:
@gtas-verisk, Thanks for filing the issue.

I think removing an added item should not place it in the internal _deleted collection.

[...]

PR #137, opened 2016-08-30
Adding Removed M-M Item Adds It to _deleted Collection

tonysneed:
This PR includes failing test from PR #136 from @MarcDrexler, which shows that calling `GetChanges` on employee.Territories with a `_deleted` item adds it back to the Territories collection.

The probable cause of this behavior is that removing an Added territory sets its state to Unchanged and adds it to the `_deleted` collection, as described in issue #132.

The problem is in the `RemoveItem` method of `ChangeTrackingCollection`, where items are added to `_deleted` if their state is not set to Added.  Because the removed territory has had its state set to Unchanged, it is added to `_deleted`.  Setting the state to Unchanged is correct for M-M relations, so we need to capture that its original state was Added, and then **not** add it to `_deleted`.

The code in question is [here](https://github.com/TrackableEntities/trackable-entities/blob/develop/Source/TrackableEntities.Client/ChangeTrackingCollection.cs#L243-L246):

```csharp
// Cache deleted item if not added or already cached
if (item.TrackingState != TrackingState.Added
    && !_deletedEntities.Contains(item))
    _deletedEntities.Add(item);
```

### DIFF
--- a/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
@@ -152,6 +152,24 @@ namespace TrackableEntities.Client.Tests
             Assert.Equal(TrackingState.Added, employee.TrackingState);
             Assert.True(employee.Territories.All(t => t.TrackingState == TrackingState.Added));
         }
+        
+        [Fact]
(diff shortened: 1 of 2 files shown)


tonysneed:
This PR fixes #132: After removing an item that has Added state it goes to the _deleted internal collection with state Unchanged.

PR #140, opened 2016-08-31
Fix #139: ChangeTrackingCollection is modified when calling GetChanges

### LINKED ISSUE #139: ChangeTrackingCollection is modified when calling GetChanges
Following Scenario:
An employee has 3 territories in its territory collection and I remove all of them.
The territory collection count is zero.

After calling GetChanges, the territory collection is not empty anymore and shows 1 territory with tracking state "Deleted". 

The collection should not be modified when calling GetChanges.

See failing test in #138

tonysneed:
The problem is with the part of `TrackableExtensions.RemoveRestoredDeletes`, which attempts to remove deletes which were temporarily restored:

```csharp
var removedDeletes = changeTracker.GetChanges(true).Cast<ITrackable>().ToList();
if (item.TrackingState == TrackingState.Deleted)
{
    var isTracking = changeTracker.Tracking;
    changeTracker.Tracking = false;
    if (removedDeletes.Contains(item))
        items.Remove(item);
    changeTracker.Tracking = isTracking;
}
```

Calling `changeTracker.GetChanges` again here seems to be the source of the problem.  Instead the `items.RemoveAt(i)` should be called.

### DIFF
--- a/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
@@ -641,6 +641,26 @@ namespace TrackableEntities.Client.Tests
             Assert.Equal(TrackingState.Modified, changes.ElementAt(0).TrackingState);
             Assert.Equal(TrackingState.Deleted, changes.ElementAt(2).TrackingState);
         }
+        
+        [Fact]
+        public void GetChanges_Should_Not_Change_Territory_Collection_Of_Tracked_Employee()
+        {
+            // Arrange
+            var database = new MockNorthwind();
+            var employee = database.Employees[0];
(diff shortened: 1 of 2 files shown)


tonysneed:
While calling changeTracker.GetChanges seems senseless and should be refactored, the real issue is with this code:

```csharp
// Prevent endless recursion
if (!visitationHelper.TryVisit(item)) continue;
```

The problem is that, in M-M relations, the visitationHelper will have visited both sides of the relationship.  So the continue statement will be executed and the item will not be removed.

tonysneed:
I changed the code for removing the deletes as follows:

```csharp
if (item.TrackingState == TrackingState.Deleted)
{
    var isTracking = changeTracker.Tracking;
    changeTracker.Tracking = false;
    items.RemoveAt(i);
    changeTracker.Tracking = isTracking;
}
```

Then I made sure this code gets executed even if the object has been visited.

PR #141, opened 2016-08-31
Calling GetChanges multiple times does not capture additional changes

MarcDrexler:
I found another one 😄 

Assuming somebody calls GetChanges several times to check if the data is dirty, the return value of GetChanges will become incorrect.

After removing 3 territories from an employee, GetChanges should return 3 territories with TrackingState Deleted, no matter how often GetChanges has been called before.

The following line in the new test will fail:
`Assert.Equal(3, changes[0].Territories.Count); `

I think you can reopen #139

### DIFF
--- a/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
@@ -660,6 +660,37 @@ namespace TrackableEntities.Client.Tests
 
             // Assert
             Assert.Equal(0, employee.Territories.Count);
+            Assert.Equal(3, changes[0].Territories.Count);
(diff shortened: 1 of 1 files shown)


tonysneed:
Fixed this bug with #143.  @MarcDrexler, great job finding bugs -- keep em coming!

tonysneed:
@MarcDrexler, By the way .. Are you using Visual Studio 2015 with Git?  There is an extension for GitHub, and the Team Explorer is great for a lot of Git tasks.  Creating branches is a breeze. 😄

MarcDrexler:
Yes, I am using VS 2015 with GitHub extension. But I am still not familiar with all the Git tasks (I'm coming from the tfs with none or less branches, so I need to change my thinking). But I will try to create a branch the next time. 😄

tonysneed:
Great to get up to speed on Git - no way I could go back to TF version control.

PR #148, opened 2016-09-28
UnitOfWork with RepositoryFactory

noufionline:
Unit of work which accepts repository factory


noufionline:
This pull requested is superseded with #149

PR #159, opened 2017-05-04
added Test Apply_Changes_Should_Mark_Added_Territories_And_Same_Remov…

edi0177:
Added this test-case to describe the issue i'am facing with N-M Entities.

This is a simple example of my Use-Case.

I try to describe it in the northwind-shema.
Let's say i have a List of Employees.
And I have a List of Territories.

So I want to remove one Territory from one employee and add the same Territory to another employee.

But the Territory-Object is in both cases not the same, it can't be the same because one is in trackingState.Added and the other one in deleted.

Having this in object-graph leads to the 

> System.InvalidOperationException: Attaching an entity of type 'TrackableEntities.EF.Tests.NorthwindModels.Territory' failed because another entity of the same type already has the same primary key value. This can happen when using the 'Attach' method or setting the state of an entity to 'Unchanged' or 'Modified' if any entities in the graph have conflicting key values. This may be because some entities are new and have not yet received database-generated key values. In this case use the 'Add' method or the 'Added' entity state to track the graph and then set the state of non-new entities to 'Unchanged' or 'Modified' as appropriate.

because the Entity is already Attached to the Context

Exception comes from  SetEntityState

` // Set state normally if we cannot perform interception
            if (interceptors == null)
            {
                context.Entry(item).State = state; <-- This fails if a other Territory with same id is already attached
                return;
            }`

My usecase is a little bit more complex but in short this is what i facing.
My Question is how to solve this or is this  (maybe unwanted behavior ) in ApplyChanges and if we should maybe check if the entity is already attached and in State unchanged and if so to decide to not set the State again.

### DIFF
--- a/Source/Tests/TrackableEntities.EF.5.Tests/NorthwindDbContextTests.cs
+++ b/Source/Tests/TrackableEntities.EF.5.Tests/NorthwindDbContextTests.cs
@@ -1152,7 +1152,7 @@ namespace TrackableEntities.EF5.Tests
 			var territory3 = employee.Territories[2];
 			var territory4 = nw.Territories[3];
 			territory4.TrackingState = TrackingState.Added;
-			employee.Territories.Add(territory4);
+            employee.Territories.Add(territory4);
 
 			// Act
 			context.ApplyChanges(employee);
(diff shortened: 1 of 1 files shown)


tonysneed:
@edi0177 Sorry I hadn't looked at this .. I'll investigate and resolve.  Let me know if you have any other insights.

PR #167, opened 2017-06-23
Added shared entities doc

weitzhandler:
Added a page that provides a brief summary of the steps any project needs to implement so *Trackable Entities* works properly with it.

### DIFF
--- a/2d-gettingstarted-shared-entities.md
+++ b/2d-gettingstarted-shared-entities.md
@@ -1,0 +1,31 @@
+---
+layout: default
+title: Getting Started - Adding Trackable Entities to an existing project
+home: true
+---
+
+This sums up the requirements a project has to comply to for **Trackable Entities** to work properly.
+
+## Declaration
+### 1. Server entities
+ - Implement `ITrackable`
+ - Implement `IMergeable`
+
+### 2. Client entities
+ - Inherit from `EntityBase`
+ - All collection reference properties must be an instance of `ChangeTrackingCollection` (CTC).* All navigation properties must internally wrap the navigated property in a singular CTC as demonstrated [here](https://github.com/TrackableEntities/trackable-entities/blob/develop/Samples/VS2013/WebApiSample.Inheritance/SampleWebApi.Inheritance.Shared.Entities/Models/Product.cs#L83)
+
+##Usage
+### 1. Server (controller etc.)
+ - Call `ApplyChanges` ex. method on provided entity argument
+ - Call the `DbContext`'s `SaveChanges(Async)`
+ - Call `LoadRelatedEntitiesAsync` ex. method to load updated stuff
+ - Call `AcceptChanges` to flatten the changed data and mark it unchanged
+ - Return a success code along with the updated entity  
+
+### 2. Client
+ - Wrap root entity/s in a CTC
+ - If using JSON, serializing using `TypeNameHandling.Auto` in settings
+ - Call the CTC's `GetChanges` method which returns a collection containing the root entity/s - if there are changes
+ - Determine if the collection contains any elements, and send to server
+ - Upon a successful result, call the local CTC's `MergeChanges` passing the server-returned updated entity
--- a/2-gettingstarted.md
+++ b/2-gettingstarted.md
@@ -21,3 +21,7 @@ Step-by-step tutorial for using Trackable Entities with **Repository and Unit of
 Step-by-step tutorial for using Trackable Entities with **Windows Communication Foundation**.
 
 [Windows Communication Foundation](2c-gettingstarted-wcf.html)
+
+## Existing projects / General requirements of Trackable Entities
+A brief summary of the requirements a projects has to comply to for *Trackable Entities* to work properly.
+[Shared Entities](2d-gettingstarted-shared-entities.html)


2d-gettingstarted-shared-entities.md
```diff
@@ -0,0 +1,31 @@
+---
+layout: default
+title: Getting Started - Adding Trackable Entities to an existing project
+home: true
+---
+
+This sums up the requirements a project has to comply to for **Trackable Entities** to work properly.
+
+## Declaration
+### 1. Server entities
```
tonysneed: Provide some code as an example.

2d-gettingstarted-shared-entities.md
```diff
@@ -0,0 +1,31 @@
...
+ - Implement `ITrackable`
+ - Implement `IMergeable`
+
+### 2. Client entities
+ - Inherit from `EntityBase`
+ - All collection reference properties must be an instance of `ChangeTrackingCollection` (CTC).* All navigation properties must internally wrap the navigated property in a singular CTC as demonstrated [here](https://github.com/TrackableEntities/trackable-entities/blob/develop/Samples/VS2013/WebApiSample.Inheritance/SampleWebApi.Inheritance.Shared.Entities/Models/Product.cs#L83)
```
tonysneed: Provide sample in code fence (triple back ticks). Use the WebApiSample project.

tonysneed: Use the models in the [portable](https://github.com/TrackableEntities/trackable-entities/blob/develop/Samples/VS2013/WebApiSample/WebApiSample.Shared.Entities.Portable/Models/Product.cs) shared folder.

tonysneed: Note that the code in M-M relations is a bit different. The [Parent](https://github.com/TrackableEntities/trackable-entities/blob/develop/Samples/VS2013/WebApiSample/WebApiSample.Shared.Entities.Portable/Models/Employee.cs#L103) property needs to be set on both sides.

tonysneed: Here is the code on the [Territory.Employees](https://github.com/TrackableEntities/trackable-entities/blob/develop/Samples/VS2013/WebApiSample/WebApiSample.Shared.Entities.Portable/Models/Territory.cs#L43) property.

2d-gettingstarted-shared-entities.md
```diff
@@ -0,0 +1,31 @@
...
+
+##Usage
```
tonysneed: This Usage section doesn't belong here in this doc, because it describes general usage versus customizing entities.

weitzhandler: I would then rather change the title of the page to something like "TE Implementation check-list". I think it does belong here. My request is there should be a short summary of all programmatic requirements needed to be implemented for TE to work in any project, no matter if it's shared, generated or custom entities.

tonysneed: I think what we need to focus on here is simply how to customize EF code first entities as trackable entities that are shared between client and server.  So you should rename the file: **2d-gettingstarted-cf-dharted-entities.md**.  Stick to this one topic in this file.

The usage of these entities on the **server** is already described in the other Getting Started pages.  The server-side will usage will vary depending on whether you are using Web API, Patterns, or WCF.

The usage of the entities on the **client** will be the same for *all* scenarios, no matter the kind of entities (CF, separate, shared), so I would suggest creating a separate file called: **2e-gettingstarted-client.md**, in which is described how to use the client-side API.  You can pretty much copy this from the **Retrieving Entities** section of the [Web API Getting Started](http://trackableentities.github.io/2a-gettingstarted-webapi.html).  Then we can update the other getting started sections to link to this client-side getting started.

weitzhandler: Oh I get it now

2d-gettingstarted-shared-entities.md
weitzhandler: In the portable example, I don't understand why [this](https://github.com/TrackableEntities/trackable-entities/blob/develop/Samples/VS2013/WebApiSample/WebApiSample.Shared.Entities.Portable/Models/Product.cs#L94) line actually does, the `ChangeTrackingCollection` paramless ctor makes the CTC a non-trackable one, so what's the purpose of wrapping it?

tonysneed:
Yes, change trackers for navigation properties start in a non-tracked state. They are set to a tracked state by the parent change tracker when they it has been set to a tracked state. Does that make sense?

weitzhandler:
No, I don't understand, who is setting their `Parent` property? How are they discovered?

tonysneed:
Change trackers use reflection to walk down an object graph and can communicate with change trackers down the graph, including both collection and reference properties. So if you set the Tracking property on an object at the top of the graph, all the trackers down the object graph are set to the same value (true or false).

There are unit tests for this behavior which you can find in the source code. You might want to check them out. Unit tests in he TE project document expected behaviors, and you debut them, stepping into the TE libraries source code.

weitzhandler:
Meaning this prevents dynamic trackers.
Have you seen [my code](https://gist.github.com/weitzhandler/117ad8f6d10ab6b997362fe5a8930500)? Is there a way to make this work?

tonysneed:
Forgive me if you've already said this somewhere, but can you please explain what you're trying to accomplish and what you mean by "dynamic trackers"?

weitzhandler:
I don't want to manually include private fields of wrapped navigation properties.
I want them to be automatically implemented.
In shared entities, you can't override the properties.

tonysneed:
> Meaning this prevents dynamic trackers.

Can you explain what you mean by this? How would hierarchical change trackers prevent dynamic trackers?

weitzhandler:
I want to apply TE to an existing model.
How can I apply it on one-to-one entities?

Consider the following 1-0..1 model. How do I enable TE without re-implementing its properties?
I thought CTC is self-controlling itself, so I made a dictionary wrapping all reference properties, as you can see [here](https://gist.github.com/weitzhandler/117ad8f6d10ab6b997362fe5a8930500#file-appentitybase-cs-L51), but if they're retrieved by reflection, how can inject those CTCs in the model automatically?

    class Order
    {
      public int Id { get; set; }
      //[ForeignKey(nameof(OrderDetail))]
      //public int? OrderDetailId { get; set; }
      public virtual OrderDetail { get; set; }
    }
    class OrderDetail
    {
      [Key, ForerignKey(nameof(Order))]
      public int OrderId { get; set; }
      public Order Order { get; set; }
    }

tonysneed:
OK, I finally see what you're trying to do now. Sorry I didn't catch onto this earlier. 😳

The way change tracking currently works on the client is that it expects change trackers to be part of the entity object graph, both as the collection type and attached to reference properties.  The reason is so that it can reason about relationship types on the client without the need for extensive metadata about the model.  This is because client entities are supposed to be agnostic about persistence concerns and should not have anything like model attributes, which are specific to frameworks like EF.

For this reason you need to have those change trackers in the client entities in a way that reflects the model and all its relationships (1-1, 1-M, M-1, M-M).  If you want to take an existing model and apply change tracking to it on the client, the way to go about that is with partial classes, in which you place all the change trackers.  But this is still a lot of work with large models, and you usually need the help of some kind of code generator such as T4.  That is why I recommended using Code First on the server to generate the database, then use T4 templates to generate the shared entities.  That is by far the easiest way to get the kinds of entities that support change tracking across service boundaries.

Now if I had it to do all over again, I would take a different approach.  I created a repo with a [wiki](https://github.com/TrackableEntities/TrackableEntities.Express.Home/wiki/design-goals) that has design goals laid out for a new version of TE that supports POCO entities on the client (paragraph no. 2), without the need for a CTC or anything like that.  This means starting from scratch on building out a new change tracker that can accept hints on the model structure.  Needless to say, this represents a significant time commitment and also a team of developers to share the effort.

The reason I haven't invested the required effort is that there is already a solution out there that does pretty much the same thing as TE -- [Breeze](http://www.getbreezenow.com) from IdeaBlade.  This is probably the reason why my application to the .NET Foundation didn't even receive a response from MS, aside from the fact that its goal was to replace a defunct solution from the EF team called Self-Tracking Entities.

So in your case you might want to have a look at Breeze to see if it fits your needs. But I do appreciate your interest in TE and questions -- it got me thinking about the project again!

weitzhandler:
Breeze does looks promising, but suffers from the same issue. No support for shared entities.

weitzhandler:
But I actually found a simple workaround to that, I'm using Fody's BeforeAfter feature to call Breeze's protected `SetValue` which subsequently raises the `PropertyChanged` event.

tonysneed:
@weitzhandler Glad you found a solution.  All the best in your endeavors!

weitzhandler:
I'm afraid I haven't found a solution yet. After reading Breezes docs and spending a nice amount of time experimenting with it, turns out it's [not compatible with Xamarin](https://github.com/Breeze/breeze.sharp/issues/22#issue-239329373).

So back to TEs, is there a way I can make it work? It's been [3 weeks](https://github.com/aspnet/EntityFramework/issues/8744#issuecomment-306559924) and I'm still in this mud...
I'm really curious what other devs do!

tonysneed:
Are there a large number of entities in your model?

weitzhandler:
BTW, I started adopting TE because you [told me](https://github.com/aspnet/EntityFramework/issues/8744#issuecomment-306607963) I can stick to my shared entities and T4 isn't mandatory...

>Are there a large number of entities in your model?

No. About 30 entities.
I chose to use shared entities, because I have validation that is supposed to undergo both on client and server.

tonysneed:
Follow me on Twitter, and we can arrange a chat via Skype. I have a bit more time now. My Twitter: tonysneed

tonysneed:
The answer might be fluent validation, which allows you to place validation login in separate classes.

weitzhandler:
I've already based my validation in DataAnnotations, so before I switch to generated entities and fluent validation, I would just like to confirm a few things:

- How is TEs determining the reference fields' CTC (I mean [this](https://github.com/TrackableEntities/trackable-entities/blob/develop/Samples/VS2013/WebApiSample/WebApiSample.Shared.Entities.Portable/Models/Product.cs#L99) field), reflection?
- Is inheritance supported by TE

tonysneed:
> Yes by reflection, and yes inheritance is supported.

In terms of requiring T4, what I meant is that you don't have to use T4 to generate the entities, but you have to make those entities conform to certain requirements, so that essentially you end up with shared entities that look like what T4 would have generated.

If you want to stick with data annotations you can start with shared entities that look like what T4 generates (or reverse engineer the entities using T4), then apply the annotations to those entities.

weitzhandler:
I now see that you actually left this part open via [`INavigationPropertyInspector`](https://github.com/TrackableEntities/trackable-entities/blob/master/Source/TrackableEntities.Common/TrackableExtensions.cs#L85), so what I'm gonna do is have my [`AppBaseEntity`](https://gist.github.com/weitzhandler/117ad8f6d10ab6b997362fe5a8930500) implement this interface.

weitzhandler:
I still don't understand why you need a private CTC wrapping the *-0..1 references.

weitzhandler:
I now see that you actually left this part open via [`INavigationPropertyInspector`](https://github.com/TrackableEntities/trackable-entities/blob/master/Source/TrackableEntities.Common/TrackableExtensions.cs#L85), so what I'm gonna do is have my [`AppBaseEntity`](https://gist.github.com/weitzhandler/117ad8f6d10ab6b997362fe5a8930500) implement this interface.

I still don't understand why you need a private CTC wrapping the *-0..1 references.
Why doesn't it work in my scenario? The 1-1/M-1 references are exposed with by their EntityBase-subclass type. Why isn't it recognized as reference?

tonysneed:
It's to detect a change on an entity on the other side of the relationship. It's not the only possible approach, but it's the one TE uses.

weitzhandler:
How does it detect changes if it's initialized as non-tracking (parameterless constructor)?

tonysneed:
It starts listening for changes as soon as it's Tracking property is set to True, by a higher change tracker in the graph.

weitzhandler:
But those CTCs are private, who knows about them?

weitzhandler:
My goal is to automate all the actions implemented by the T4 templates in the `AppEntityBase`.

weitzhandler:
What does [my code](https://gist.github.com/weitzhandler/117ad8f6d10ab6b997362fe5a8930500) miss that is not reverse-engineering the T4 templates behavior?

tonysneed:
Via reflection it's possible to walk the object graph and read private properties.

tonysneed:
> My goal is to automate all the actions implemented by the T4 templates in the AppEntityBase.
> 
There are hundreds of tests validating the client change tracker, and you would have to make sure they all pass with your base class.

That sounds like a lot more work than simply creating 30 entities like the ones generated by the T4 templates.  Why don't you just create the entities using T4, then add your validation attributes?

weitzhandler:
[Here](https://github.com/TrackableEntities/trackable-entities/blob/master/Source/TrackableEntities.Common/TrackableExtensions.cs#L223) you call the [`GetProperties`](https://msdn.microsoft.com/en-us/library/aky14axb) method that only loads the public properties.

tonysneed:
Have you run the unit tests locally? You would have to create another set of tests using your base class instead of EntityBase.

tonysneed:
@weitzhandler Did you still want to do a Skype chat? I think I might be able to help you find a solution more easily if we speak in real time. Just send me a tweet and I can direct message you my Skype id. You can also email me: tony@tonysneed.com.

weitzhandler:
Can we have the [`CloneObject`](https://github.com/TrackableEntities/trackable-entities/blob/develop/Source/TrackableEntities.Client/TrackableExtensions.cs#L274) aksi rely on the `IRefPropertyChangeTrackerResolver` interface, so if the entity implements it, it serializes the deleted entities along.

tonysneed:
Closing this as we've discussed these items and opened other PR's to resolve.

PR #168, opened 2017-07-11
Entity properties marked NotMapped should not be marked as modified

weitzhandler:
Hi,

1. Entity properties marked as `NotMapped`, shouldn't be `entry.Property(prop).IsModified` marked upon.
2. Verify all properties in `ModifiedProperties` actually exist in the entity, they might be removed during serialization, or added manually.

PR #170, opened 2017-07-13
Ignore non-mapped properties from ModifiedProperties

weitzhandler:
Addressing #168 

Properties attributed `NotMapped` in `ModifiedProperties` should be excluded from `ApplyChanges`.

TODO: Not mapped properties by Fluent API.


Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -207,8 +208,16 @@ public static void ApplyChanges(this InterceptorPool pool, IEnumerable<ITrackabl
                 {
                     // Mark modified properties
                     SetEntityState(context, item, parent, propertyName, EntityState.Unchanged, interceptors);
-                    foreach (var property in item.ModifiedProperties)
-                        context.Entry(item).Property(property).IsModified = true;
+                    var entry = context.Entry(item);
+                    // Exclude properties attributed as NotMapped
+                    var notMappedProps = item.GetType()
```
weitzhandler: I really believe the solution should be Edm-based rather than reflection-based, but that's what I've come up with as an immediate solution.

tonysneed:
Thanks for putting this together .. I'll start by running the failing test.  Then I'll have a look at your fix.

Source/Tests/TrackableEntities.EF.5.Tests/FamilyModels/Parent.cs
```diff
@@ -13,6 +13,9 @@ public Parent(string name)
 
         [Key]
         public string Name { get; set; }
+
+        public string Nickname { get; set; }
```
weitzhandler: Ignore this commit, I forgot adding the `NotMapped` attr.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
tonysneed: Yes, if this is handled on the server-side, it will have to be EDM-based, because non-mapped properties can be set via fluent API.

If this is too difficult, I would suggest instead doing this on the client-side, with an extension method that looks for the [NotMapped] attribute using reflection and then adds those properties to `ExcludedProperties` on the CTC.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -225,7 +234,7 @@ public static void ApplyChanges(this InterceptorPool pool, IEnumerable<ITrackabl
                 else
                 {
                     context.ApplyChangesOnProperties(item, visitationHelper.Clone(), TrackingState.Modified, interceptors); // Clone to avoid interference
-                    context.ApplyChangesOnProperties(item, visitationHelper, TrackingState.Deleted, interceptors);
```
tonysneed: @weitzhandler Make sure you don't commit changes to things like spacing and indentation, so those files won't appear in the diff.  That will help me focus on the changes that really matter. :)

Source/Tests/TrackableEntities.EF.5.Tests/FamilyModels/Parent.cs
tonysneed: Here's a git hint: If you notice you forgot to add something to the previous commit.  You can make the change and simply [amend the last commit](https://help.github.com/articles/changing-a-commit-message/).

Don't be afraid to do a `git push --force` of your branch.  Because it's a PR, it won't affect anything else.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
weitzhandler: It sure belongs to server because `NotMapped` is not even supported on all clients. So it has to be skipped in the server side.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
weitzhandler: ok

Source/TrackableEntities.EF.5/DbContextExtensions.cs
tonysneed: OK, so see if you can make this EDM based.  You need to write a private method called `IsMappedProperty(string propertyName)` and call it like so:

```csharp
// Set modified properties
if (item.TrackingState == TrackingState.Modified
    && (state == null || state == TrackingState.Modified)
    && item.ModifiedProperties != null
    && item.ModifiedProperties.Count > 0
    && !IsMappedProperty(propertyName)
)
```

Put it in the `ApplyChanges Helpers` region of `DbContextExtensions`.  You can make use of the `GetEdmSpaceType` method.  I think the `DeclaredProperties` of `EntityType` should exclude the `NonMapped` (data annotations) and `Ignored` (fluent API) properties, but I'm not sure.

Hopefully this helps! :)

weitzhandler: Oh I missed this comment, but I implemented it without a separate method, I think it should be sufficient. Check my last commit please dd99154e2beb3584d938d78d11f856ce78340344.

tonysneed:
Having a look at this ...

tonysneed:
Looks good .. I'll merge this in.

tonysneed:
@weitzhandler Do you know how to do a squash commit?  Use interactive rebase to squash all your commits into a single commit, then run `git push --force` to submit it.  Notice how your prior commits are replaced with the squashed commit.  Then I can just merge in that commit.  Thanks in advance!

weitzhandler:
I'm not using git cli, only git extension in VS.

tonysneed:
@weitzhandler Looks like changes here conflict with PR #177.  I'll try to do that, and also the squash commit.

weitzhandler:
The changes from the `Not-Mapped` patch should be discarded.

tonysneed:
Thanks.  I'll process this.

tonysneed:
@weitzhandler I'm seeing a failing test: `Apply_Changes_Should_Skip_NotMapped_Properties`
System.InvalidOperationException
Member 'IsModified' cannot be called for property 'Nickname' on entity of type 'Parent' because the property is not part of the Entity Data Model.

This [code](https://github.com/TrackableEntities/trackable-entities/blob/develop/Source/TrackableEntities.EF.5/DbContextExtensions.cs#L215) is throwing the exception: `property.IsModified = true;`  There needs to be a check if the property is not mapped.  Can you implement this?  Thanks!

weitzhandler:
I get this error in my code too.
Working on it now.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -359,7 +359,9 @@ where GetEntityTypes(dbContext, entityType).Contains(es.ElementType)
             var selectedItems = loadAll ? items
                 : items.Where(t => t.TrackingState == TrackingState.Added
                     || (parent != null && parent.TrackingState == TrackingState.Added));
-            var entities = selectedItems.Cast<object>().Where(i => !IsComplexType(context, i.GetType()) && !visitationHelper.IsVisited(i));
+
+            var entities = selectedItems.Cast<object>()
+                .Where(i => !IsComplexType(context, i.GetType()) && !visitationHelper.IsVisited(i));
```
weitzhandler: Just to comply with the develop branch. Nothing major, please ignore.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -523,6 +525,17 @@ private static bool IsComplexType(this DbContext dbContext, Type entityType)
...
 
+        private static bool IsMappedProperty(this DbContext dbContext, Type entityType, string propertyName)
+        {
+            MetadataWorkspace workspace = ((IObjectContextAdapter)dbContext)
+                .ObjectContext.MetadataWorkspace;
+
+            StructuralType oType = workspace.GetItems<StructuralType>(DataSpace.OSpace)
+                .Where(e => e.FullName == entityType.FullName).SingleOrDefault();
+
+            return oType.Members.Select(member => member.Name).Contains(propertyName);
```
weitzhandler: The four lines in this method, are identical to the IsComplexType and another method, shouldn't we integrate between them?

weitzhandler:
All issues fixed.
Please see also #182

weitzhandler:
Migrated to #192.

Issue #171, opened 2017-07-14
Feature request: Introduce new GraphChanged event


weitzhandler:
Would be nice to have an event that raises when any property change or collection change occurs within the entire graph becomes dirty.

The use case of this is implementing an auto-save that is executed upon a change in the graph (probably with a timer).

What I'm using meanwhile is a timer that periodically checks for changes, but we could optimize that by running the timer only if there are changes to allow a grace period delay of finishing string input etc.

The `EventArgs` should provide data about the changed object and its parent, and the change that occurred.

tonysneed:
What you really need is a dirty checker.  It should take a snapshot of a graph and be able to detect a change. It's out of scope really for TE, but I bet a modern MVVM framework will have it. Maybe MVVM Cross? Or maybe someone wrote a dirty checker and published to NuGet?

weitzhandler:
The checker exists. I need an event. So if the user dreams for 20 secs the timer shouldn't have to take snapshots all of that time.

weitzhandler:
A dirty checker is built in TE (`GetChanges().Any()`) already, but I can only figure out if the graph's dirty by having a timer that checks this every few seconds.
An event would allow for turning the save delay timer only if the user actually made it dirty, otherwise the save timer will be idle too.
A dirty count (i.e. number of total modification in the entire graph) could also help (less optimal tho), so that the continuous timer runs every few seconds, and runs a `DelaySave` method, which turns on a delay timer, that checks the dirty count. If it's the same as it was few seconds ago, trigger the actual save, otherwise allow user to make more changes before saving.

But really what we need is an event, so that when user changes text anywhere in the graph, it resets the save delay timer. Can you please point out extension points I can implement to achieve that?

tonysneed:
I think there is an EntityChanged event, or something like that?

Also, see about HasChanges extension method.

weitzhandler:
The `EntityChanged` event only tracks current entity.
Where should I add/remove listeners by walking through the graph?

weitzhandler:
The `EntityChanged` event only tracks current entity.
Where should I add/remove listeners by walking through the graph?

Would you accept a PR?

weitzhandler:
The `EntityChanged` event only tracks current entity.
Where should I add/remove listeners by walking through the graph?

Would you accept a PR?
[...]

PR #173, opened 2017-07-15
GraphChanged event

weitzhandler:
Addresses #171

### DIFF
--- a/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
@@ -719,6 +719,23 @@ namespace TrackableEntities.Client.Tests
             Assert.Equal(3, changesCount);
         }
 
+        [Fact]
+        public void GraphChanged_Event_Should_Fire_When_Graph_Items_Added_Modified_Deleted()
+        {
+          //Arrange
+          var database = new MockNorthwind();
+          var product = database.Products[0];
+          var changeTracker = new ChangeTrackingCollection<Product>(product);
+          var raised = false;
+          changeTracker.GraphChanged += (s, e) => raised = true;
+
+          //Act
+          product.Category.CategoryName = "Foo";
+
+          //Assert
+          Assert.True(raised);
+        }        
+
         #endregion
-    }
+  }
 }
--- a/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
@@ -734,10 +734,11 @@ namespace TrackableEntities.Client.Tests
           // Act
           order.OrderDetails.Add(new OrderDetail());
           order.Customer.CustomerName = "Tony sneed";
-          order.OrderDetails.RemoveAt(0);   
+          order.OrderDetails.RemoveAt(0);
+          order.OrderDetails[0].Product.Category.CategoryName = "Lorem ipsum";
 
           // Assert
-          Assert.Equal(3, changeCount);
+          Assert.Equal(4, changeCount);
         }        
 
         #endregion
--- a/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
@@ -720,26 +720,24 @@ namespace TrackableEntities.Client.Tests
         }
 
         [Fact]
-        public void GraphChanged_Event_Should_Fire_When_Graph_Items_Added_Modified_Deleted()
+        public void EntityChanged_Event_Should_Fire_When_Graph_Items_Added_Modified_Deleted()
         {
           // Arrange
+          var changeCount = 0;
+
           var database = new MockNorthwind();
           var order = database.Orders[0];
-          var changeTracker = new ChangeTrackingCollection<Order>(order);
-          var raiseCount = 0;
-          changeTracker.EntityChanged += (s, e) => raiseCount++;
+          var changeTracker = new ChangeTrackingCollection<Order> { order };
+          changeTracker.EntityChanged += (s, e) => changeCount++;
+          changeTracker.Tracking = true;
 
           // Act
-
-          // Added
           order.OrderDetails.Add(new OrderDetail());
-          // Modified
           order.Customer.CustomerName = "Tony sneed";
-          // Deleted
           order.OrderDetails.RemoveAt(0);   
 
           // Assert
-          Assert.Equal(3, raiseCount);
+          Assert.Equal(3, changeCount);
         }        
 
         #endregion
--- a/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
@@ -722,18 +722,24 @@ namespace TrackableEntities.Client.Tests
         [Fact]
         public void GraphChanged_Event_Should_Fire_When_Graph_Items_Added_Modified_Deleted()
         {
-          //Arrange
+          // Arrange
           var database = new MockNorthwind();
-          var product = database.Products[0];
-          var changeTracker = new ChangeTrackingCollection<Product>(product);
-          var raised = false;
-          changeTracker.GraphChanged += (s, e) => raised = true;
+          var order = database.Orders[0];
+          var changeTracker = new ChangeTrackingCollection<Order>(order);
+          var raiseCount = 0;
+          changeTracker.EntityChanged += (s, e) => raiseCount++;
 
-          //Act
-          product.Category.CategoryName = "Foo";
+          // Act
 
-          //Assert
-          Assert.True(raised);
+          // Added
+          order.OrderDetails.Add(new OrderDetail());
+          // Modified
+          order.Customer.CustomerName = "Tony sneed";
+          // Deleted
+          order.OrderDetails.RemoveAt(0);   
+
+          // Assert
+          Assert.Equal(3, raiseCount);
         }        
 
         #endregion
--- a/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
@@ -732,13 +732,11 @@ namespace TrackableEntities.Client.Tests
           changeTracker.Tracking = true;
 
           // Act
-          order.OrderDetails.Add(new OrderDetail());
-          order.Customer.CustomerName = "Tony sneed";
-          order.OrderDetails.RemoveAt(0);
+          order.OrderDetails[0].Product.Discontinued = !order.OrderDetails[0].Product.Discontinued;
           order.OrderDetails[0].Product.Category.CategoryName = "Lorem ipsum";
 
           // Assert
-          Assert.Equal(4, changeCount);
+          Assert.Equal(2, changeCount);
         }        
 
         #endregion
--- a/Source/TrackableEntities.Client/ChangeTrackingCollection.cs
+++ b/Source/TrackableEntities.Client/ChangeTrackingCollection.cs
@@ -21,16 +21,31 @@ namespace TrackableEntities.Client
     {
         // Deleted entities cache
         private readonly Collection<TEntity> _deletedEntities = new Collection<TEntity>();
-
+        
+        //private event EventHandler _EntityChanged;
         /// <summary>
         /// Event for when an entity in the collection graph has changed its tracking state.
         /// </summary>
         public event EventHandler EntityChanged;
+    //{
+    //    add
+    //    {
+    //        _EntityChanged += value;
+    //        // TODO avoid recursive call
+    //        SetTracking(Tracking, new ObjectVisitationHelper(), false, _EntityChanged);
+    //    }
+    //    remove
+    //    {
+    //        _EntityChanged -= value;
+    //        // TODO avoid recursive call
+    //        SetTracking(Tracking, new ObjectVisitationHelper(), false, _EntityChanged);         
+    //    }
+    //}
 
-        /// <summary>
-        /// Default contstructor with change-tracking disabled
-        /// </summary>
-        public ChangeTrackingCollection() : this(false) { }
+    /// <summary>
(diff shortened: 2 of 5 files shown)


Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
```diff
@@ -719,6 +719,23 @@ public void EntityChanged_Event_Should_Fire_When_Items_Added_Modified_Deleted()
...
 
+        [Fact]
+        public void GraphChanged_Event_Should_Fire_When_Graph_Items_Added_Modified_Deleted()
+        {
+          //Arrange
+          var database = new MockNorthwind();
+          var product = database.Products[0];
+          var changeTracker = new ChangeTrackingCollection<Product>(product);
+          var raised = false;
+          changeTracker.GraphChanged += (s, e) => raised = true;
```
tonysneed: I would prefer to keep EntityChanged, which fires when *any* entity in a graph of tracked objects has changed.

weitzhandler: What do you mean? Should we keep only one event that fires when the graph is changed?

weitzhandler:
Reading [your comment](https://github.com/TrackableEntities/trackable-entities/issues/171#issuecomment-315548991), if I'm to implement this in the `SetTracking` ex. method, since the event is in the CTC:
1. How do I refer to the CTC from the `SetTracking` method, is `visitationHelper.OfType<ITrackingCollection>().First()` reliable?
2. Should I create a method (maybe internal) `RaiseEntityChanged` so that the event can be raised from the ex. method (events can only be raised in current class)?

tonysneed:
Concerning subscribing to `EntityChanged`: There is also a `SetTracking` method in the CTC, which works in coordination with `SetTracking` in TrackableExtensions. The one in CTC can reference `this.EntityChanged`, and the one in TrackableExtensions references both the ref and collection property change trackers.

I don't think you have to worry about raising `EntityChanged`, because it is already being raised whenever an entity property is modified and when the entity is added or removed.

weitzhandler:
Yes but what are my options to acquire the children.
As I see it, I have two options:

1. Pass a reference to the CTC via an argument in the `SetTracking` ex. method, and pass on all descending CTC's `EntityChanged` event to parent (via a `public`/`internal` `RaiseEntityChanged` method on the CTC
2. Have the `SetTracking` ex. method return an `IEnumerable<ITrackingCollection>` of all processed entities for the `CTC.SetTracking` to subscribe to their `EntityChanged` event.

I hope you have better ideas.

tonysneed:
OK it looks like we need to change the signature of `TrackableExtensions.SetTracking` to accept an additional parameter: `EventHandler entityChanged`.  Then when a CTC reference is encountered, add it to the `EntityChanged` event:

```csharp
refChangeTracker.EntityChanged += entityChanged;
colProp.EntityCollection.EntityChanged += entityChanged;
```

Then in `ChangeTrackingCollection.SetTracking`:

```csharp
item.SetTracking(value, visitationHelper, oneToManyOnly, EntityChanged);
```

I don't know for sure, but I think this will result in the EntityChanged event of the topmost CTC getting fired whenever an entity is changed anywhere in the object graph.

tonysneed:
PS. You need to do `+=` when tracking is being turned on, and `-=` when it is being turned off.

weitzhandler:
The problem with this way is that it will work only if the handlers are attached before setting the tracking. We have to raise the event, not fire the handlers (which might join the party in a later stage thus neglected).

weitzhandler:
The problem with this way is that it will work only if the handlers are attached before setting the tracking. We have to raise the event, not fire the handlers (which might join the party in a later stage hence will be
 neglected).

This will not work:

    //handler is null when SetTracking is called
    var changeTracker = new ChangeTrackingCollection<Order>(order);

    //handler wasn't attached to graph entities
    changeTracker.EntityChanged += handler;

~~~Maybe we should expand the `EntityChanged` event, so when a handler is add/removed, it calls `SetTracking` all over again, or a different method that only attaches/unattaches the handlers without touching the tracking. Not sure if even possible. Anyway yikes.~~~

tonysneed:
Can you create a commit with your attempt and then push it? Then I can pull and try it out. Thanks!

Source/TrackableEntities.Client/ChangeTrackingCollection.cs
```diff
@@ -21,16 +21,31 @@ public class ChangeTrackingCollection<TEntity> : ObservableCollection<TEntity>,
...
         // Deleted entities cache
         private readonly Collection<TEntity> _deletedEntities = new Collection<TEntity>();
-
+        
+        //private event EventHandler _EntityChanged;
         /// <summary>
         /// Event for when an entity in the collection graph has changed its tracking state.
         /// </summary>
         public event EventHandler EntityChanged;
+    //{
```
weitzhandler: Notice this.

tonysneed:
I'll check it out ...

Source/TrackableEntities.Client/ChangeTrackingCollection.cs
```diff
@@ -90,15 +105,15 @@ public bool Tracking
             }
             set
             {
-                SetTracking(value, new ObjectVisitationHelper(), false);
+                SetTracking(value, new ObjectVisitationHelper(), false, EntityChanged);
```
tonysneed: Don't pass `EntityChanged` here.

Source/TrackableEntities.Client/ChangeTrackingCollection.cs
```diff
@@ -90,15 +105,15 @@ public bool Tracking
...
             }
         }
 
         /// <summary>
         /// For internal use.
         /// Turn change-tracking on and off with proper circular reference checking.
         /// </summary>
-        public void SetTracking(bool value, ObjectVisitationHelper visitationHelper, bool oneToManyOnly)
+        public void SetTracking(bool value, ObjectVisitationHelper visitationHelper, bool oneToManyOnly, EventHandler entityChanged)
```
tonysneed: Don't add `entityChanged` to this method.

Source/TrackableEntities.Client/ITrackingCollection.cs
```diff
@@ -27,8 +27,8 @@ public interface ITrackingCollection : ICollection
         /// <summary>
         /// For internal use.
         /// </summary>
-        void SetTracking(bool value, Common.ObjectVisitationHelper visitationHelper,
-            bool oneToManyOnly);
+        void SetTracking(bool value, Common.ObjectVisitationHelper visitationHelper, 
+            bool oneToManyOnly, EventHandler entityChanged);
```
tonysneed: Don't change the interface.

Source/TrackableEntities.Client/TrackableExtensions.cs
```diff
@@ -37,7 +42,8 @@ internal static class TrackableExtensions
                         if (refChangeTracker != null)
                         {
                             // Set tracking on ref prop change tracker
-                            refChangeTracker.SetTracking(enableTracking, visitationHelper, oneToManyOnly);
+                            refChangeTracker.SetTracking(enableTracking, visitationHelper, oneToManyOnly, entityChanged);
+                            refChangeTracker.SetHandler(enableTracking, entityChanged);
                         }
```
tonysneed: Just call `SetHandler` here.

Source/TrackableEntities.Client/TrackableExtensions.cs
```diff
@@ -47,9 +53,22 @@ internal static class TrackableExtensions
                 {
                     bool isOneToMany = !IsManyToManyChildCollection(colProp.EntityCollection);
                     if (!oneToManyOnly || isOneToMany)
-                        colProp.EntityCollection.SetTracking(enableTracking, visitationHelper, oneToManyOnly);
+                    {
+                        colProp.EntityCollection.SetTracking(enableTracking, visitationHelper, oneToManyOnly, entityChanged);
+                        colProp.EntityCollection.SetHandler(enableTracking, entityChanged);
+                    }
```
tonysneed: Don't pass `entityChanged` to `SetTracking.` Just call `SetHandler` here.

Source/TrackableEntities.Client/TrackableExtensions.cs
```diff
@@ -60,7 +79,7 @@ internal static class TrackableExtensions
         /// <param name="visitationHelper">Circular reference checking helper</param>
         /// <param name="isManyToManyItem">True is an item is treated as part of an M-M collection</param>
         public static void SetState(this ITrackable item, TrackingState state, ObjectVisitationHelper visitationHelper,
-            bool isManyToManyItem = false)
+                bool isManyToManyItem = false)
         {
             ObjectVisitationHelper.EnsureCreated(ref visitationHelper);
```
tonysneed: The visitor should prevent recursion.  We can circle back to take a closer look at this.

tonysneed:
OK, so I made some comments on the code.  Let me know if they make sense .. I haven't implemented them, so they may or may not be valid. :)  It should not matter when the handlers are attached.

weitzhandler:
1. You didn't say anything about my [previous comment](https://github.com/TrackableEntities/trackable-entities/pull/173#issuecomment-315617759)
2. Not changing the pattern in the CTC (only in the ex. method) will only pass the handler on to the first layer of children, but not to the navigation properties (which is turned on by calling their own CTC's `SetTracking`. I'll improve the test to cover that.

tonysneed:
> Not changing the pattern in the CTC (only in the ex. method) will only pass the handler on to the first layer of children, but not to the navigation properties

`Set Tracking` in `TrackableExtensions` walks down the graph getting all the downstream CTC's.  The idea is that the handlers are chained.  Maybe the best way is for me to implement the way I'm describing and share it with you in a PR.  I could be wrong, but let me try that .  It's easier for me to demonstrate than just theorizing. :)

weitzhandler:
The `SetTracking` ex. method, calls the nav. properties' `SetTracking`, which is the `CTC`'s, not the ex. method, thus, the parent `CTC`'s `EntityChanged` is neglected.

tonysneed:
> The SetTracking ex. method, calls the nav. properties' SetTracking of the CTC

That's not how it's supposed to work .. let me demonstrate that.  I'll do a PR that shows what I mean ...

weitzhandler:
You can use the same [branch](https://github.com/weitzhandler/trackable-entities/tree/graph-changed-event)

tonysneed:
OK, so you gave me push rights?

weitzhandler:
Yes

tonysneed:
Can you resend the invite?  I clicked decline by mistake. ;)

weitzhandler:
Done

tonysneed:
Working on this, but I have to put the kids to bed first. :)

weitzhandler:
Haha, mine are sleeping tight, they're soon to get up (hope not too soon)...

Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
```diff
@@ -719,6 +719,28 @@ public void EntityChanged_Event_Should_Fire_When_Items_Added_Modified_Deleted()
...
+          var changeCount = 0;
+
+          var database = new MockNorthwind();
+          var order = database.Orders[0];
+          var changeTracker = new ChangeTrackingCollection<Order> { order };
+          changeTracker.EntityChanged += (s, e) => changeCount++;
+          changeTracker.Tracking = true;
+
+          // Act
+          order.OrderDetails.Add(new OrderDetail());
```
tonysneed: So add this code to test the event on the parent:

```csharp
order.OrderDate = order.OrderDate.AddDays(1);
```

Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
```diff
@@ -719,6 +719,28 @@ public void EntityChanged_Event_Should_Fire_When_Items_Added_Modified_Deleted()
...
+          order.Customer.CustomerName = "Tony sneed";
+          order.OrderDetails.RemoveAt(0);
+          order.OrderDetails[0].Product.Category.CategoryName = "Lorem ipsum";
```
tonysneed: Just need to figure out why change to `Category` on line 738 does not fire the EntityChanged event.  See if we're not subscribing to it.

Source/TrackableEntities.Client/TrackableExtensions.cs
tonysneed: Ignore my previous comment here .. looks like the only change was in the spacing.  Can you please discard the spacing changes? :-)

tonysneed:
It looks like I've run out of time on this for now, and I won't be able to work on it myself for a while.  So I think the next step here would be to add the [line](https://github.com/TrackableEntities/trackable-entities/pull/173#discussion_r127701071) to the test that covers the root entity.  That will determine if the root entity is included.

Then you can figure out why the [change](https://github.com/TrackableEntities/trackable-entities/pull/173#discussion_r127701493) to `Category.CategoryName` is not firing the event.

weitzhandler:
I'd like to get this one too.
Very important to me having an even raised whenever anything changes in the entire graph.

tonysneed:
Thanks for working on this .. I'll plan on getting it into the 2.5.6 release.

tonysneed:
I'm getting a failing test on this: `EntityChanged_Event_Should_Fire_When_Graph_Items_Added_Modified_Deleted`
Xunit.Sdk.EqualException
Assert.Equal() Failure
Expected: 4
Actual:   3

Maybe `CategoryChangeTracker` is not tracking changes?

weitzhandler:
I'm afraid it's not enough changing the signature in the extensions, we also need to change the signature of `SetTracking` in the CTC to accept an handler for the child references when called recursively.

That's why these two don't raise the event:

    order.OrderDetails[0].Product.Discontinued = !order.OrderDetails[0].Product.Discontinued;
    order.OrderDetails[0].Product.Category.CategoryName = "Lorem ipsum";

tonysneed:
I'll have a look.

tonysneed:
Can you try implementing your proposal, and we can see if it works?

weitzhandler:
How would you prefer implementing it? Should I change the existing `SetTracking` method in the interface? Add a new one? I now you wouldn't like changing the interface, that's why I wanted to introduce it as a separate `GraphChanged` event.

tonysneed:
I figured it out .. have a look at PR #186 Recursive EntityChanged event in ChangeTrackingCollection.

PR #174, opened 2017-07-16
Failing test for providing deleting reference properties

weitzhandler:
We should allow reference properties' container entities to carry their old values, and when needed, be serialized back to server.
I'd like to see the `*ChangeTracker` properties vanish in favor of using an `IDictionary<string, ITrackable> DeletedReferences`, which will be part of the common `ITrackable`, and will return the deleted children of the current entity.

The default should be to exclude the deleted ref. properties, but we should give option to setup dependent/independent props.
For example `Product.Category` is independent, because `Category` is standalone regardless of the `Product`s using it, while `Customer.Address` is dependent, meaning when I set `Customer.Address` to a new address, I'd want the old `Address` entity to be deleted. And its deletion might have to be smart because it might carry a foreign key of the `Customer` which can be the same as the `Added` entity, so we first need to delete old, save than add `ApplyChanges` and save again.

### DIFF
--- a/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
+++ b/Source/Tests/TrackableEntities.Client.Tests/ChangeTrackingCollectionTests.cs
@@ -577,7 +577,7 @@ namespace TrackableEntities.Client.Tests
             var database = new MockNorthwind();
             var changeTracker = new ChangeTrackingCollection<Product>(true);
             var product = database.Products[0];
-            changeTracker.Add(product);
+            changeTracker.Add(product);    
 
             // Act
             var changes = changeTracker.GetChanges();
@@ -616,6 +616,25 @@ namespace TrackableEntities.Client.Tests
 
             // Assert
             Assert.Equal(TrackingState.Deleted, changes.First().TrackingState);
+        }
+
+        [Fact]
+        public void GetChanges_Should_Return_Deleted_References()
+        {
+            // Arrange
+            var database = new MockNorthwind();
+            var product = database.Products[0];
+            var category = product.Category;
+            var changeTracker = new ChangeTrackingCollection<Product>(product);
+
+            // Act
+            product.Category = new Category();
+
+            // Assert
+            Assert.Equal(TrackingState.Added, product.Category.TrackingState);
+            // We should add another property in the ITrackable interface that returns the deleted interface.
+            // Maybe as a dictionary specifying property name
+            Assert.True(false); //product.DeletedReferences[nameof(Product.Category)].Contains(category);
         }
 
         [Fact]


tonysneed:
> We should allow reference properties' container entities to carry their old values, and when needed, be serialized back to server.

Why?

weitzhandler:
So server knows to delete them. We could also do this manually on server, but it's easier when server gets them and decides what to do with them.

tonysneed:
Sorry, I must be missing something here.  If an entity is marked as deleted, the server will delete it.  Can you give an example?

weitzhandler:
public class Customer : EntityBase { 
      public int Id { get; set; }
      public Address Address { get; set; }
    }
    public class Address : EntityBase {
      public int Id { get; set; }
    }

After loading `Customer` `1` that has `Address` `1`, I set `Customer`'s `Address` to a new `Address`, the old `Address` should be passed to the server, so the server deletes it.

tonysneed:
Ah I see now.  It looks like you're missing the for foreign key in Customer.  The way TE works is that you need to use foreign keys.  The reason for this is a) performance (you're not needlessly sending entities across the wire), and b) forward compat with EF vNext, which is requiring foreign key relations.

tonysneed:
```csharp
public class Customer : EntityBase { 
  public int Id { get; set; }
  public int AddressId { get; set; }
  public Address Address { get; set; } // this will be null when posted to the server
}
```

tonysneed:
Changing `AddressId` on `Customer` should not necessarily require that the old address be deleted.  However, if that is your intent, you need to simply pass an `Address` entity marked as deleted and call `ApplyChanges` on the server. The deletion will be part of the same transaction when `SaveChanges` is called.

weitzhandler:
I omitted it here for brevity, my issue still applies.
I want, that when the `Customer.Address` property is `null`ed or changed in the client side, the old `Address` instance should be provided to the server as `Deleted`.

weitzhandler:
> if that is your intent, you need to simply pass an Address entity marked as deleted and call ApplyChanges on the server.

What if the old entity and the new entity carry the same `Id` (in case of foreign key)?

tonysneed:
You're talking about a M-1 relation?  For example, `Order.Customer` in Northwind?

weitzhandler:
No, talking about N-0..1.

tonysneed:
Wouldn't this work?

```csharp
// send both Customer and Address to the server
public class Customer { 
  public int Id { get; set; }
  public int AddressId { get; set; } // set to new id
  public Address Address { get; set; } // set to null
}
public class Address { // marked as deleted
  public int Id { get; set; }
}
```

weitzhandler:
No.
Setting `Address` to null will add `Address` to `Customer`'s `ModifiedProperties`. The old `Address` however is ignore, and doesn't appear in the `DeletedEntities` either.

weitzhandler:
We should:
1. Change the private properties convention to a public dictionary that appears in both ends (client/server)
2. Deleted references (ref: N-1..0, col = N..M) should thus be extracted from the CTC maintaining its parent via the `DeletedEntities` property - which is not.

tonysneed:
> Setting Address to null will add Address to Customer's ModifiedProperties

That's not  right.  The `GetChanges` method sets it to null.

weitzhandler:
But it does so after it's extracted the restored the deleted entities.
I'm requesting the same behavior for refs, not only collections.

tonysneed:
> The old Address however is ignore, and doesn't appear in the DeletedEntities either.

Yes, there is a problem here, because you need to pass the address you want to delete separately.  Right now, the way you would do that is to pass in the entity separately.

weitzhandler:
I already made my hacks in my code (but then I bumped into the conventional `*ChangeTracker` private properties), but it woulda been easier if it was managed by the system automatically. Because in data binding, the UI is bound directly to the property, and without a `PropertyChanging` event, it's cumbersome getting the old references.
Anyway we should get rid of the named private properties and convert them into a serializable property that exists in both ends and will always serialize (if not empty).

tonysneed:
If you mark a ref entity as deleted, then it will be returned by `GetChanges`.  So what I think is happening is that you need to perform two operations.  You need to indicate that the foreign key value has changed, and also that the old foreign entities is to be deleted.

It seems to me that `GetChanges` should do this.  And then `ApplyChanges` on the server side need to know to inform EF that the entity is to be deleted.

The problem I have in general is making TE a bit "too smart."  There are some cases where you will need to mark entities manually, or when you need to pass more than one root entity to a service and call `ApplyChanges` on each one.  The concern is that if we try to accommodate all the edge cases, it's going to require some work to implement, debug and maintain.

On the other hand, if you're willing to do the work, then we can give it a try.  So if you're game, then we would need updates to both `GetChanges` and `ApplyChanges`.  But my caution is that this may require a non-trivial amount of work. :)

weitzhandler:
> If you mark a ref entity as deleted, then it will be returned by GetChanges

But if you null it, no. This test fails:

    [Fact]
    public void Test()
    {
      var db = new MockNorthwind();
      var prod = db.Products[0];
      var ct = new ChangeTrackingCollection<Product>(prod);
      prod.Category = null;
      var changes = ct.GetChanges();
      Assert.Equal(TrackingState.Deleted, changes.Single().Category?.TrackingState);
    }

> The problem...

First thing we need to do, in order to get the serialization cover up everything, is get rid of the conventional names & private properties. It should all be transfered via a public property exposed in `ITrackable`, do you agree?

tonysneed:
I'm not sure what you mean by cover up everything .. what is not being transferred (serialized) that should be?

weitzhandler:
There should be a way to be able to transfer the deleted entities of ref props. Right now, they'll not be included in the serialization, unless stored before changing the property in the containing property.
Even set as deleted, as soon as we set the ref property with a new instance, the deleted entity isn't tracked anymore.
A ref property should be wrapped in a CTC capable of tracking two entities: the original one and the new one (if any).
By default the old entity shouldn't be marked as `Deleted` thus shouldn't be included in `GetChanges` but if set `Deleted` (or if we find a nice way to setup the CTC to recognize dependency of entities and the related entity is a dependent one), it should be included in the serialization.

tonysneed:
> There should be a way to be able to transfer the deleted entities of ref props. Right now, they'll not be included in the serialization, unless stored before changing the property in the containing property.

If you mark a reference property entity as deleted it will be included when `GetChanges` is called.  This is probably one of those cases where you may need to manually set TrackingState to Deleted on those entities, because setting the reference property to null does not result in the the change tracker marking them as Deleted.  This presently is just a limitation of how change tracking works on the client.  But the same thing holds true I think of EF's change tracker.

> Even set as deleted, as soon as we set the ref property with a new instance, the deleted entity isn't tracked anymore.

Yes, and this is why you need a separate change tracker for the deleted entity.  The rules as they now stand (and as far as I can remember at this point) is that reference entities are not marked as deleted by the client change tracker of a root change tracker.  And this is as it should be.  In fact, it's the same behavior as the EF change tracker, which also requires you to independently mark an entity as deleted in order for EF to delete it.

So for example, you have `Order` with a `Customer` property (M-1).  The root change tracker is for the Order.  You can set CustomerId to null (if it's nullable) or change it.  But that does not mean the Customer should be deleted.  In fact, most of the time referential constraints will prevent it, if other orders have the same CustomerId.  If you want to delete the customer, you need a CTC of Customer, and you can remove it or manually mark it as deleted.

> ... but if set Deleted (or if we find a nice way to setup the CTC to recognize dependency of entities and the related entity is a dependent one), it should be included in the serialization.

This is where I do not want TE to behave differently than how change tracking works in EF.  If you need to delete a Customer (M-1 related to Order), then you need to have a Delete operation on the server that accepts a Customer.  Semantically, I don't think deleting the customer should be part of the same transaction as updating an order, because of possible relational constraint violations.

Sorry if this seems too limiting.  But it's just some of the design decisions we've had to make early on, which has aligned TE with EF in terms of change tracking, transactions, foreign keys, concurrency, etc.

The good news is that for your domain, there are ways you can get around this limitation if you want to.  For example, you can have a service operation that accepts an Order root entity, and also a Customer(s) root entity(ies).  Then you call `ApplyChanges` once for each of these.  When `SaveChanges` is called, they will both be included in the same transaction.

weitzhandler:
I understand, and thanks for the detailed explanation!

OK, regarding the private tracker properties. In my domain I have 4 levels of entities, including inheritance on each of them. It's pretty cumbersome deleting all the references and track them individually.
I've already managed to set it all up and test it (for deletion of ref-properties).
The only thing I don't manage to implement as an external patch is serializing the properties. The serialization should rely on the `IRefPropertyChangeTrackerResolver` interface, not on private properties.
So in order to provide those reference in serialization, I suggested having an additional property in the entity.

tonysneed:
> In order to provide those reference in serialization, I suggested having an additional property in the entity.

So you're referring to a `DeletedEntities` property ... I seem to recall you had a base class extending `EntityBase`, maybe `AppEntityBase`?  If that's right, why don't you put `DeletedEntities` in that class?  Of course, this would be in your project, not TE, but I think it's reasonable to extend TE in this manner.  What do you think?

weitzhandler:
That's what I actually did...
But the internal clone/serialization methods seem to skip it...

tonysneed:
> But the internal clone/serialization methods seem to skip it...

Ah .. there is an extensibility point which should allow you to bring it in .. I just can't remember offhand (I think someone on the team implemented that at one point).  Maybe you can investigate?

weitzhandler:
I hope i'll find time.
If you look thru the proj and it does pop at you, I'd appreciate a hint!

tonysneed:
Yes, I should be able to have a look in a couple of hours from now.

tonysneed:
OK, so what you need to do is add a test for cloning an entity that has a property of type `Dictionary<string, ITrackable>`.  `Clone` uses a JSON serializer, which should serialize the dictionary.  I would also suggest placing a `[JsonProperty]` attribute on the property to see if that helps.

tonysneed:
@weitzhandler I added a passing test for cloning a `Dictionary<string, ITrackable>` in an entity base class.  Seems to work alright. See PR #176.

weitzhandler:
Thank you Tony.
Will calling `GetChanges` still persists those dictionaries? How about omitting unchanged entities?

tonysneed:
> Will calling GetChanges still persists those dictionaries?

Yes, it should.

> How about omitting unchanged entities?

Unchanged entities will be omitted as expected.

tonysneed:
@weitzhandler Can I close this, since serialization of dictionaries seems to work (see #176)?

weitzhandler:
@tonysneed 
Do you ever consider adding `DeleteReferences` dictionary to `ITrackable`? It will be very helpful to provide M-0..1 references that were deleted, as a an addition and replacement for the conventional private properties.

tonysneed:
Better to create another interface to contain a Dictionary .. single responsibility principle.

weitzhandler:
Sounds good.

weitzhandler:
Sounds good.
Should we have it implement both `ITrackable` and `Imergeable` or again, single responsibility?

tonysneed:
Just a single responsibility .. Also part of SOLID .. the "I" is for Interface Segregation Principle.

weitzhandler:
👍

tonysneed:
Going to close this, so that a new PR can be opened to extend TE's ability to include composite graphs of entities.

PR #177, opened 2017-07-20
ComplexType support

weitzhandler:
Let's implement support for complex types in TrackableEntities.

Currently calling `ApplyChanges` method when there is a `ComplexType` in the graph, raises the following exception:
>System.InvalidOperationException: 'The type 'MyComplexType' is mapped as a complex type. The Set method, DbSet objects, and DbEntityEntry objects can only be used with entity types, not complex types.'

StackTrace:
>at System.Data.Entity.Internal.InternalContext.UpdateEntitySetMappingsForType(Type entityType)


Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -199,16 +199,31 @@ public static void ApplyChanges(this InterceptorPool pool, IEnumerable<ITrackabl
...
+                        if(isComplex)
+                        {
+                            var parentEntry = context.Entry(parent);
+                            if(parentEntry.State == EntityState.Detached)
+                              parentEntry.State = parent.TrackingState.ToEntityState();
+                            parentEntry.ComplexProperty(propertyName).Property(property).IsModified = true;                           
+                        }
+                        else
+                            context.Entry(item).Property(property).IsModified = true;
+                    }
```
weitzhandler: I thought about moving the `context.Entry()` call outside the `for each` loop so we don't have to retrieve the entity again on each iteration.
The ugly thing here is that `parentEntry.ComplexProperty(propertyName)` returns a `DbComplexPropertyEntry` and `context.Entry(item)`, a `DbEntityEntry<ITrackable>` which are not assignable from each other, so the `Property()` function is not a shared one. Meh, could do it if I hadn't involved the complex properties. Apparently `ComplexProperties` do come with their tiny complications.

tonysneed:
Looks like you're on the right track. Have to go to bed just now, but I'll pull and look more closely in the morning my time. Cheers.

Source/Tests/TrackableEntities.EF.5.Tests/Contexts/FamilyDbContext.cs
```diff
@@ -32,7 +32,15 @@ public FamilyDbContext(CreateDbOptions createDbOptions = CreateDbOptions.CreateD
...
         }
 
-        public DbSet<Parent> Parents { get; set; }
+#if EF_6
+        protected override void OnModelCreating(DbModelBuilder modelBuilder)
+        {
+            foreach (var property in typeof(ITrackable).GetProperties())
+              modelBuilder.Types().Configure(m => m.Ignore(property.Name));
+        }
+#endif
```
tonysneed: Why do we need to ignore ITrackable members?  Those should be excluded already.

Source/Tests/TrackableEntities.EF.5.Tests/FamilyDbContextTests.cs
```diff
@@ -215,5 +215,21 @@ public void Apply_Changes_Should_Mark_Grandchild_Deleted()
...
         }
+
+        [Fact]
+        public void Apply_Changes_Should_Mark_ComplexType_Modified()
+        {
+            // Arrange
+            var context = TestsHelper.CreateFamilyDbContext(CreateFamilyDbOptions);
+            var parent = new MockFamily().Parent;              
+            parent.Address.StreetName = "123 Lee Ave.";
+            parent.Address.ModifiedProperties = new[] { nameof(Address.StreetName) };
```
tonysneed: What's missing here is to set `TrackingState` to `Modified`.  AFAIK currently TE first checks `TrackingState` before `ModifiedProperties`.  The reason is that the purpose of `ModifiedProperties` is to allow for partial updates.  If there are no modified properties, then a full update us performed if `TrackingState` is `Modified`.  So to stay consistent, I'd like to keep this requirement for complex types as well.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
tonysneed: Actually, `DbComplexPropertyEntry` inherits from `DbPropertyEntry`, so you could clean this up be adding a getPropertyEntry method that returns `DbPropertyEntry` and setting `IsModified` to true.

```csharp
DbPropertyEntry GetPropertyEntry(...)
// if complex type return DbComplexPropertyEntry
```

Source/Tests/TrackableEntities.EF.5.Tests/FamilyDbContextTests.cs
tonysneed: So change the code in `DbContextExtensions` so that this test fails.  Then go back and change the test to add:

```csharp
parent.Address.TrackingState = TrackingState.Modified
```

Then the test will pass.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -496,9 +512,21 @@ private static EntityType GetEdmSpaceType(this DbContext dbContext, Type entityT
...
+                .ObjectContext.MetadataWorkspace;
+
+            StructuralType oType = workspace.GetItems<StructuralType>(DataSpace.OSpace)
+                .Where(e => e.FullName == entityType.FullName).SingleOrDefault();
+
+            return oType.BuiltInTypeKind == BuiltInTypeKind.ComplexType;
+        }
+
+
+    #region ApplyChanges Helpers
```
tonysneed: This is good. 👍

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -199,16 +199,31 @@ public static void ApplyChanges(this InterceptorPool pool, IEnumerable<ITrackabl
                     context.ApplyChangesOnProperties(item, visitationHelper.Clone(), TrackingState.Deleted, interceptors);
                 }
 
+                var isComplex = context.IsComplexType(item.GetType());
+                bool isParentEligible(TrackingState parentState) => parentState == TrackingState.Modified || parentState == TrackingState.Unchanged;
+                var isModifiable = 
+                          (isComplex 
+                            ? isParentEligible(parent.TrackingState)
```
tonysneed: Need to check TrackingState is Modified for complex types as well.  See next comments for suggestion to simplify this a bit.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -199,16 +199,31 @@ public static void ApplyChanges(this InterceptorPool pool, IEnumerable<ITrackabl
...
+                            ? isParentEligible(parent.TrackingState) 
+                            : (item.TrackingState == TrackingState.Modified && (state == null || state == TrackingState.Modified)))                          
+                          && item.ModifiedProperties?.Count > 0;
+
                 // Set modified properties
-                if (item.TrackingState == TrackingState.Modified
-                    && (state == null || state == TrackingState.Modified)
-                    && item.ModifiedProperties != null
-                    && item.ModifiedProperties.Count > 0)
+                if (isModifiable)
```
tonysneed: I would suggest creating an `IsModifiable(ITrackable item)`  method to encapsulate all the logic.

```csharp
private bool IsModifiable(ITrackable item) {
    if (item.TrackingState == TrackingState.Modified
        && (state == null || state == TrackingState.Modified)
        && item.ModifiedProperties != null
        && item.ModifiedProperties.Count > 0) {
            if (!context.IsComplexType(item.GetType())) return true;
        return parent.TrackingState == TrackingState.Modified || parent.TrackingState == TrackingState.Unchanged;
    }
    return false;
}
```

Check the syntax, but I think this will do the trick.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -724,6 +752,9 @@ private static RelationshipType GetRelationshipType(this DbContext dbContext, Ty
             ITrackable item, ITrackable parent, string propertyName,
             EntityState state, IEnumerable<IStateInterceptor> interceptors)
         {
+
+            if (context.IsComplexType(item.GetType())) return;
```
tonysneed: This is fine.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
weitzhandler: I've fixed it in a slightly different way that saves the same purpose.

Source/Tests/TrackableEntities.EF.5.Tests/Contexts/FamilyDbContext.cs
weitzhandler: OK undone

Source/Tests/TrackableEntities.EF.5.Tests/FamilyDbContextTests.cs
weitzhandler: I think with CTs, we don't really care about the state of the CT, but rather about its parent state, no?

Source/TrackableEntities.EF.5/DbContextExtensions.cs
weitzhandler: I actually created a method already, then I thought I wouldn't want to pollute your class with more method so I stuck it all inline, I'm glad you're also voting for extraction. Done.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
weitzhandler: Done.

weitzhandler:
@tonysneed 
> If you agree, go ahead and implement my suggested changes. Then we can see why this test is reporting 27 changes instead of 26: Accept_Changes_Should_Mark_Family_Unchanged()

After addressing all reviews that test is still failing. See my new commit.

tonysneed:
Isn't the 27th change coming from Parent.Address? If so, then we change the Assert in the test.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -199,16 +199,21 @@ public static void ApplyChanges(this InterceptorPool pool, IEnumerable<ITrackabl
...
-                    && item.ModifiedProperties != null
-                    && item.ModifiedProperties.Count > 0)
+                if (isModifiable(isComplex))
                 {
                     // Mark modified properties
                     SetEntityState(context, item, parent, propertyName, EntityState.Unchanged, interceptors);
-                    foreach (var property in item.ModifiedProperties)
-                        context.Entry(item).Property(property).IsModified = true;
+                    var entry = isComplex ? context.Entry(parent) : context.Entry(item);
+                    if(isComplex && entry.State == EntityState.Detached) entry.State = parent.TrackingState.ToEntityState();
```
tonysneed: I'm wondering about this line, which is new.  What if the parent is unchanged and the CT is modified?
Also, why check if EntityState is Detached?  We're not doing this anywhere else in `ApplyChanges`.

weitzhandler: A CT cannot be attached, its parent is.

Source/Tests/TrackableEntities.EF.5.Tests/FamilyDbContextTests.cs
weitzhandler: Done.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -243,8 +233,19 @@ public static void ApplyChanges(this InterceptorPool pool, IEnumerable<ITrackabl
                     context.ApplyChangesOnProperties(item, visitationHelper, TrackingState.Deleted, interceptors); 
                 }
             }
-        }
 
+        bool isModifiable(bool isComplex)
+        {
+          if (item.TrackingState == TrackingState.Modified
+              && (state == null || state == TrackingState.Modified)
+              && item.ModifiedProperties?.Count > 0)
```
weitzhandler: Can also be changed to:

    item.ModifiedProperties?.Any() == true

Essentially the same, just wasting another call doing it via LINQ.

tonysneed: Let's leave these lines the same as the are now.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
tonysneed: OK, so I commented out line [210](https://github.com/TrackableEntities/trackable-entities/commit/f5df58c264e626d414cc3c968af429d8b7900f61#diff-18c35e3f0eae888189d043a03dbb3ef2R210), and all the unit tests still pass.  If you think this line is necessary, can you create a failing test that will pass when the line is inserted?

tonysneed: Otherwise, we can leave the line out, and the PR is ready to be merged.

weitzhandler: Awesome.
I believe it was also related to setting the CT's state as `Modified`.
So yes we should delete this line unless we'll have trouble in the future. After attaching my real-life project to the updated TE version I'll automatically have more test cases, hopefully no trouble is encountered.

weitzhandler:
Hooray!

weitzhandler:
Is it already on of the feeds? Any plans?

weitzhandler:
What's missing is, since complex type are not entities, I get the following exception when calling `LoadRelatedEntitiesAsync`:

NullReferenceException: 
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source)
   at TrackableEntities.EF6.DbContextExtensions.IdMatcher.GetKeyValue(Object entity)
   at TrackableEntities.EF6.DbContextExtensions.IdMatcher.System.Collections.Generic.IEqualityComparer<System.Object>.GetHashCode(Object obj)
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   at System.Collections.Generic.Dictionary`2.ContainsKey(TKey key)
   at TrackableEntities.Common.ObjectVisitationHelper.IsVisited(Object obj)
   at TrackableEntities.Common.ObjectVisitationHelper.TryVisit(Object obj)
   at TrackableEntities.EF6.DbContextExtensions.<LoadRelatedEntitiesAsync>d__2d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

Will fix the error and commit.

PR #190, opened 2017-08-03
Added NPC into the main BaseEntity

### LINKED ISSUE #161: NotifyPropertyChanged
@tonysneed 
Please include another overload that allows raising the `PropertyChanged` event in `EntityBase`. Currently only a lambda expression is accepted, allowing no direct way to raise property change by string from sub classes.

Besides, there is also the `CallerMemberNameAttribute` that can be solved this way. Will you accept a PR?

Temporary workaround: reflection 😢 

    protected void RaisePropertyChanged([CallerMemberName] string propertyName = null)
    {
      var eventArgs = new PropertyChangedEventArgs(propertyName);
      var type = GetType();
      var inpcDelegate = 
        (MulticastDelegate)type.GetRuntimeField(nameof(PropertyChanged)).GetValue(this);
      if (inpcDelegate != null)
        foreach (var handler in inpcDelegate.GetInvocationList())
          handler.GetMethodInfo().Invoke(handler.Target, new object[] { this, eventArgs });
    }

weitzhandler:
Added NPC into the main BaseEntity, so portable classes also get the `CallerMemberName` attribute if they're above .NET 4.5.

Fixes #161.

### DIFF
--- a/Source/TrackableEntities.Client/EntityBase.cs
+++ b/Source/TrackableEntities.Client/EntityBase.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.ComponentModel;
 using System.Linq.Expressions;
+using System.Runtime.CompilerServices;
 
 namespace TrackableEntities.Client
 {
@@ -23,17 +24,18 @@ namespace TrackableEntities.Client
             (Expression<Func<TResult>> property)
         {
             string propertyName = ((MemberExpression)property.Body).Member.Name;
+            // ReSharper disable once ExplicitCallerInfoArgument
             NotifyPropertyChanged(propertyName);
         }
 
-        /// <summary>
-        /// Fire PropertyChanged event.
-        /// </summary>
-        /// <param name="propertyName">Property name.</param>
+    /// <summary>
+    /// Fire PropertyChanged event.
+    /// </summary>
+    /// <param name="propertyName">Property name.</param>
 #if NET40
-        protected virtual void NotifyPropertyChanged(string propertyName)
+    protected virtual void NotifyPropertyChanged(string propertyName)
 #else
-        protected virtual void NotifyPropertyChanged([System.Runtime.CompilerServices.CallerMemberName] string propertyName = "")
+        protected virtual void NotifyPropertyChanged([CallerMemberName] string propertyName = "")
 #endif
         {
             if (PropertyChanged != null)
@@ -47,7 +49,7 @@ namespace TrackableEntities.Client
         /// </summary>
         public void SetEntityIdentifier()
         {
-            if (EntityIdentifier == default(Guid))
+            if (EntityIdentifier == Guid.Empty)
                 EntityIdentifier = Guid.NewGuid();
         }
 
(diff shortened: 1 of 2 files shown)


Source/TrackableEntities.Client/EntityBase.cs
```diff
@@ -23,11 +23,24 @@ public abstract partial class EntityBase : INotifyPropertyChanged, ITrackable, I
             (Expression<Func<TResult>> property)
         {
             string propertyName = ((MemberExpression)property.Body).Member.Name;
+            NotifyPropertyChanged(propertyName);
```
tonysneed: You can safely remove lines 25 and 26.

Source/TrackableEntities.Client/EntityBase.cs
```diff
@@ -23,11 +23,24 @@ public abstract partial class EntityBase : INotifyPropertyChanged, ITrackable, I
...
+        }
+
+        /// <summary>
+        /// Fire PropertyChanged event.
+        /// </summary>
+        /// <param name="propertyName">Property name.</param>
+#if NET40
+        protected virtual void NotifyPropertyChanged(string propertyName)
+#else
+        protected virtual void NotifyPropertyChanged([System.Runtime.CompilerServices.CallerMemberName] string propertyName = "")
```
tonysneed: Add directive: `using System.Runtime.CompilerServices;`
Update line 36:
```csharp
protected virtual void NotifyPropertyChanged([CallerMemberName] string propertyName = "")
```

Source/TrackableEntities.Client/EntityBase.cs
weitzhandler: Why would you want to do that for?

Source/TrackableEntities.Client/EntityBase.cs
weitzhandler: I thought maybe you might want to keep the namespace inline since it's conditional, but OK.

weitzhandler: I thought maybe you might want to keep the namespace inline since it's conditional, but OK.
Lemme know about the first comment so I can commit.

Source/TrackableEntities.Client/EntityBase.cs
weitzhandler: Why would you want to do that for?
I think it's best if we have a virtual `NotifyPropertyChanged` method that all calls tunnel thru it (or we add an additional empty virtual method `OnNotifyPropertyChanged` that's called just after the notification happens.
It's best to give more flexibility and option to intercept the property notification for subclasses hence making it `protected`, so we're not tied to subscribing to event and caring about memory leaks and unsubscribing.

weitzhandler: What warning is R#er giving you?

tonysneed: You can keep this code, but add the following comment after line 25:
```csharp
// ReSharper disable once ExplicitCallerInfoArgument
```

tonysneed: R# says: Explicit argument passed to parameter with caller info attribute.

weitzhandler: Ok fixed.

Source/TrackableEntities.Client/EntityBase.cs
weitzhandler: I thought maybe you might want to keep the namespace inline since it's conditional, but OK, fixed.

PR #185, opened 2017-07-25
Complex types - Finish fixing load related entities

weitzhandler:
Issue explained in [this](https://github.com/TrackableEntities/trackable-entities/pull/182#issuecomment-317878366) comments and the comments preceding it.

PR #182, opened 2017-07-24
Fixed LoadRelatedEntities

weitzhandler:
By skipping complex types.
#177

PR #191, opened 2017-08-03
Finish ComplexType

weitzhandler:
Fix another issue with LoadRelatedEntities.
Fixes #185 but now using a new branch to avoid conflicts.

Exception explained in detail in [this](https://github.com/TrackableEntities/trackable-entities/pull/182#issuecomment-317878366) comment.

### DIFF
--- a/Source/Tests/TrackableEntities.EF.5.Tests/LoadRelatedEntitiesTests.cs
+++ b/Source/Tests/TrackableEntities.EF.5.Tests/LoadRelatedEntitiesTests.cs
@@ -17,6 +17,7 @@ using System.Data.EntityClient;
 using TrackableEntities.EF.Tests;
 using TrackableEntities.EF.Tests.Mocks;
 using TrackableEntities.EF.Tests.NorthwindModels;
+using TrackableEntities.EF.Tests.FamilyModels;
 
 #if EF_6
 namespace TrackableEntities.EF6.Tests
@@ -829,6 +830,27 @@ namespace TrackableEntities.EF5.Tests
 
 #endif
 
-        #endregion
-    }
+    #endregion
+
+        #region Complex Types
+
+        [Fact]
+        public void LoadRelatedEntities_With_Complex_Types()
+        {
+            var context = TestsHelper.CreateFamilyDbContext(CreateDbOptions.DropCreateDatabaseIfModelChanges);
+            var parent = new Parent(Guid.NewGuid().ToString());
+            parent.Address = new Address();
+            context.Parents.Add(parent);
+            context.SaveChanges();
+
+            // Act
+            context.LoadRelatedEntities(parent);
+
+            // Assert
+            // previous call throws exception if not implemented properly.
+            Assert.True(true);
+        }
+
+        #endregion Complex Types
+  }
 }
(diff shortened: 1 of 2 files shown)


Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -391,9 +391,8 @@ where typeof(ITrackable).IsAssignableFrom(prop.PropertyType)
             // Recursively populate related entities on ref and child properties
             foreach (var item in items)
             {
-                // Avoid endless recursion
-                if (!visitationHelper.TryVisit(item)) continue;
-
+                // Avoid loading related entities of complext types, and avoid endless recursion
+                if (!visitationHelper.TryVisit(item) || IsComplexType(context, item.GetType())) continue;
```
tonysneed: You need to make this change also to the async version, `LoadRelatedEntitiesAsync`.

weitzhandler:
@tonysneed 
> This PR need a commit with a failing test.

Ok added a failing test and fixed it.

Source/TrackableEntities.EF.5/DbContextExtensions.cs
```diff
@@ -392,7 +392,7 @@ where typeof(ITrackable).IsAssignableFrom(prop.PropertyType)
             foreach (var item in items)
             {
                 // Avoid loading related entities of complext types, and avoid endless recursion
-                if (!visitationHelper.TryVisit(item) || IsComplexType(context, item.GetType())) continue;
+                if (IsComplexType(context, item.GetType()) || !visitationHelper.TryVisit(item)) continue;
```
tonysneed: No problem for this PR, but in the future try to add the failing test in one comment, then add the fix to another commit.  That way, I can do a soft reset to the commit with the failing test to see it fail, and then reset to the fit commit to see it pass.  Just a bit easier that way.  But don't worry about it this time.

Source/Tests/TrackableEntities.EF.5.Tests/LoadRelatedEntitiesTests.cs
```diff
@@ -829,6 +830,27 @@ public void LoadRelatedEntities_Should_Populate_Multiple_Employees_Territories_W
...
+        public void LoadRelatedEntities_With_Complex_Types()
+        {
+            var context = TestsHelper.CreateFamilyDbContext(CreateDbOptions.DropCreateDatabaseIfModelChanges);
+            var parent = new Parent(Guid.NewGuid().ToString());
+            parent.Address = new Address();
+            context.Parents.Add(parent);
+            context.SaveChanges();
+
+            // Act
+            context.LoadRelatedEntities(parent);
```
tonysneed: We'll also need a test for `LoadRelatedEntitiesAsync`.  And of course the fix for that in a separate commit.

weitzhandler: OK great, I'll now commit a failing test for the async, then a fix for it with the following commit.

weitzhandler:
Ok.
This time I committed the failing test and the fix in separate commits.
See above.

PR #194, opened 2017-08-10
EntityChanged EventArgs

weitzhandler:
Improve the event-args, so we have more info about what entity changed.


tonysneed:
Thanks! I'll have a look.

PR #200, opened 2017-11-06
Added a fix to ignore merging changes of unidentifiable entities.

### LINKED ISSUE #197: LoadRelatedEntities for 1-1 with Non-Matching Foreign Keys
Hi,

I have this graph:

```c#
Contact : Person
- Address : Address
```
(a `Contact` class with an `Address` property which is also an entity.

When calling `_collectionChangeTracker.MergeChanges(contact)`, even if the `Address` property's entity is `Unchanged`, the property is still updated with the empty data from server.
`Unchanged` entities should be ignored by the CTC.

weitzhandler:
Fixes #197.


tonysneed:
Good catch.  I'll run your test locally and then merge.

PR #202, opened 2017-11-06
Revert "Added a fix to ignore merging changes of unidentifiable entities."

tonysneed:
There are 4 failing test that need to be fixed.
- Run all the tests to see which ones are failing.


tonysneed:
Revert #200 because there are 4 failing tests, which need to be fixed.

PR #215, opened 2017-11-30
Changed ModifiedProperties default initialization to HashSet<string>

weitzhandler:
To avoid possibility of having duplicate property names.


tonysneed:
Good idea.

PR #220, opened 2018-03-28
Update ChangeTrackingCollection.cs

noufionline:
To Avoid duplicate property names to be added in ModifiedProperties Collection


noufionline:
By Instantiating Modified Collection property as HashSet<T> the problem is solved. therefore, deleting this pull request.

PR #221, opened 2018-05-17
IsComplexType throws NullReferenceException

weitzhandler:
This happens when we have a readonly property that exposes a certain type.
It makes TE think that the type is in the model while it's not.


tonysneed:
I'll have a look at this.

weitzhandler:
Thank you! Lemme know.

isakatirci:
i have same error

isakatirci:
image

tonysneed:
This fix will be available in the CI MyGet feed.

